PR #38, opened 2019-07-09
Feature/upload registration

blabadi:


src/logger.ts
```diff
@@ -0,0 +1,42 @@
...
+        format.json(),
+        timestamp(),
+        prettyPrint()
+    ),
+    transports: [
+        new winston.transports.Console()
+    ]
+};
+
+export interface Logger {
```
andricDu: :+1:

src/infra/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+            return headers.reduce<{[k: string]: string}>((obj, nextKey, index) => {
+            obj[nextKey] = data[index];
+            return obj;
+            }, {});
+        });
+        return rows.filter((el) => el != undefined);
+    };
+}
+
+export namespace ControllerUtils {
```
andricDu: It doesn't actually look like you are getting any benefit from using namespaces. 

I think their use here may actually be an anti-pattern and you should be using modules here instead or exporting the functions directly. 
https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html

src/infra/controllers/submission.ts
```diff
@@ -0,0 +1,69 @@
+import * as submission from "../../domain/services/submission";
+import { Request, Response } from "express";
+import { TsvUtils, ControllerUtils } from "../utils";
```
andricDu: A benefit of using name spacing is avoiding these sort of imports. See my other comment about module vs namespace.

PR #41, opened 2019-07-12
Feature/26 mock reg dictionary

blabadi:
add schema mock client to load the schema and store it in the db.
started foundation for validation based on the schema


blabadi:
@andricDu will do !

PR #58, opened 2019-08-02
Feature/43 auth

blabadi:
- add jwt token checks for correct scopes


tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noImplicitAny": true,
     "resolveJsonModule": true,
     "moduleResolution": "node",
+    "experimentalDecorators": true,
```
andricDu: :+1:

Issue #60, opened 2019-08-06
add endpoint to delete registration


blabadi:
- Delete the registration from MongoDB 
- this will serve the clear button in the UI

alekspejovic:
Tested in QA - lgtm!

PR #78, opened 2019-08-12
Feature/60 registration delete

jaserud:
Registration delete, swagger docs and test. Reference issue #60.


src/submission/submission-service.ts
```diff
@@ -166,6 +166,18 @@ export namespace operations {
...
 
+  /**
+   * delete registration by registration Id
+   * @param registrationId string
+   * @param programId string
+   */
+  export const deleteRegistration = async (registrationId: string, programId: string) => {
+    const registration = await registrationRepository.findById(registrationId);
+    if (registration === undefined || registration.programId !== programId) {
+      return `Cannot delete non-existent registration with id :${registrationId}`;
```
blabadi: isntead of returning string throw the error Errors.NotFound

src/submission/submission-service.ts
```diff
@@ -166,6 +166,18 @@ export namespace operations {
...
+    }
+    return await registrationRepository.delete(registrationId);
```
blabadi: no need to return this you can return true

test/integration/submission/test.data.ts
```diff
@@ -0,0 +1,27 @@
+export const ABCD_REGISTRATION_DOC = {
```
blabadi: you can put this in the same test no need for different file unless you have alot of these

src/resources/swagger.yaml
```diff
@@ -107,6 +107,31 @@ paths:
...
+          description: the short name of the program
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "401":
+          $ref: "#/components/responses/UnauthorizedError"
+        "404":
+          $ref: "#/components/responses/404NotFound"
```
blabadi: you have to add the 404NotFound I can't find it in the swagger components, use https://editor.swagger.io/ to make sure the swagger is valid

test/integration/submission/test.data.ts
blabadi: also define this as ABCD_REGISTRATION_DOC: ActiveRegistration to get type check here

PR #99, opened 2019-08-16
Update swagger.yaml

jaserud:


blabadi:
bump the version, we cant go to master without that

PR #101, opened 2019-08-16
Feature/45 lecter integration

blabadi:
call lectern with fetch, use promise tools for retry and timeout
add app health endpoint
add commit and version to be passed as env to the docker image for app health endpoint
add ego key health to health report
change schema name in env template
update health status on different events


blabadi:
@jaserud I'll merge this but please review it and see what have been added, it will affect your local environment

PR #119, opened 2019-08-22
release 0.7.1

blabadi:
Changes:
Added vault integration (vault-node)
Added user name and password env vars as secrets.
Added bunch of env var to configure vault for local testing.


blabadi:
just added you guys as fyi no need to review

PR #130, opened 2019-08-28
Feature/118 ids endpoints

blabadi:


src/auth-decorators/index.ts
```diff
@@ -67,3 +71,26 @@ export function HasSubmitionAccess(programIdExtractor: Function) {
     return descriptor;
   };
 }
+
+export function HasFullReadAccess() {
+  return function(target: any, key: string, descriptor: PropertyDescriptor) {
```
jaserud: is target used?

blabadi: no but I believe this the structure you should return to create a typescript decorator function

jaserud: ¯\_(ツ)_/¯, ok I had no idea

PR #134, opened 2019-08-29
Enhancement: Added a new endpoint /submission/template/<schemaName> t…

KevinFHartmann:
…hat returns an example tsv template for the given schemaName from the current schemaDictionary.


src/submission/template-api.ts
```diff
@@ -0,0 +1,49 @@
...
+import { SchemaDefinition, ValueType } from "../lectern-client/schema-entities";
+import { setStatus, Status } from "../app-health";
+
+const L = loggerFor(__filename);
+
+export const get = async (req: Request, res: Response) => {
+  const schemaName: string = req.params.schemaName;
+  const schemasDictionary = manager.instance().getCurrent();
+  if (!schemasDictionary) {
+    setStatus("schema", { status: Status.ERROR });
```
blabadi: I don't think this is a right place for this, I prefer we keep the status update in the bootstrap logic, I also think it will already be set to error at this point

src/submission/template-api.ts
```diff
@@ -0,0 +1,49 @@
...
+    return res.status(404).send("no schema dictionary found");
+  }
+  const schema = schemasDictionary.schemas.find((schema) => {
+    return schema.name == schemaName;
+  });
+  if (!schema) {
+    setStatus("schema", { status: Status.ERROR });
```
blabadi: same as above no need for this

src/submission/template-api.ts
```diff
@@ -0,0 +1,49 @@
...
+    return res.status(404).send("no schema named " + schemaName + " found");
```
blabadi: you can use utils.ControllerUtils.NotFound

src/submission/template-api.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+  const template = createTemplate(schema);
+  return res.status(200).send(template);
+};
+
+function createTemplate(schema: SchemaDefinition): string {
+  const header = schema.fields.map((f): string => {
+    return f.name;
+  }).join("\t") + "\n";
+  const body = schema.fields.map((f): string => {
```
blabadi: can you confirm how the template will look like with Rosi, I don't know if we need a default body line

src/app.ts
```diff
@@ -60,7 +61,8 @@ app.delete(
 
 /** Schema API */
 app.get("/submission/schema/", middleware.wrapAsync(schemaApi.get));
-
+// get template for a given schema
+app.get("/submission/template/:schemaName", middleware.wrapAsync(templateApi.get));
```
blabadi: we also have swagger file to document the endpoints, can you document this api there ? you can use editor.swagger.io to help you

src/submission/template-api.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+  const schema = schemasDictionary.schemas.find((schema) => {
+    return schema.name == schemaName;
+  });
+  if (!schema) {
+    setStatus("schema", { status: Status.ERROR });
+    return res.status(404).send("no schema named " + schemaName + " found");
+  }
+  const template = createTemplate(schema);
+  return res.status(200).send(template);
```
blabadi: I believe the gateway needs to get an actual file not a string, can you confirm that with Jon and modify

src/app.ts
blabadi: the url can be /submission/schema/templates/:schemaName and you can use the schema-api under submission instead of creating template api

src/submission/template-api.ts
KevinFHartmann: She says we don't. :-)

blabadi:
by the way this PR is pointing to master can you change it to develop

also please check our test folder to see the unit and integration tests and feel free to write tests to avoid regressions as much as possible

PR #146, opened 2019-09-16
Feature/116 validation endpoint

jaserud:


src/app.ts
```diff
@@ -62,6 +62,10 @@ app.post(
   upload.array("clinicalFiles"),
   middleware.wrapAsync(submissionAPI.saveClinicalTsvFiles)
 );
+app.post(
+  "/submission/program/:programId/clinical/validate/:versionId",
+  middleware.wrapAsync(submissionAPI.validateActiveSubmission)
+);
```
joneubank: This is fine for this PR since the whole app is setup this way, but we do need to split our endpoints into router/controller files and not have an infinitely expanding app.ts file

jaserud: agreed

PR #156, opened 2019-09-24
Feature/117 specimen validation

jaserud:
Specimen validations:
-check submitter_donor_id and submitter_specimen_id registered
-check specimen_acquisition_interval is less than donor_survival_time if donor=deceased

Donor validations:
-check submitter_donor_id registered
-check donor_survival_time is greater than all specimen_aquisition_interval if donor=deceased

Sample validations (won't work in dev or qa yet because no schema):
-check submitter_donor/specimen/sample_ids are registered

Added validation endpoint stats and pending_approval
-new/updated/noUpdate
-dataUpdates showing old and new value


src/clinical/clinical-entities.ts
```diff
@@ -5,27 +5,29 @@ export interface Donor {
   submitterId: string;
   programId: string;
   specimens: Array<Specimen>;
-  clinicalInfo?: object;
+  clinicalInfo?: { [feild: string]: string | number };
```
joneubank: field

src/clinical/clinical-entities.ts
```diff
@@ -5,27 +5,29 @@ export interface Donor {
...
+}
+
+export interface Specimen extends DonorSubEntity {
   samples: Array<Sample>;
   specimenType: string;
-  submitterId: string;
   specimenId?: number;
   tumourNormalDesignation: string;
-  clinicalInfo?: object;
+  clinicalInfo?: { [feild: string]: string | number };
```
joneubank: field

src/submission/submission-entities.ts
```diff
@@ -55,6 +55,12 @@ export type SubmissionValidationError = {
   index: number;
 };
 
+export type SubmissionValidationUpdate = {
+  fieldName: string;
+  info: object;
```
joneubank: generic object for info? is there no standard format for this?

src/submission/validation-clinical/donor.ts
```diff
@@ -1,28 +1,119 @@
...
+    };
   }
-  return errors;
+
+  // cross entity donor record validation
+  checkTimeConflictWithSpecimen(existentDonor, donorRecord, newRecords[FileType.SPECIMEN], errors);
+
+  return errors.length > 0
+    ? { type: ModificationType.ERRORSFOUND, index: donorRecord.index, resultArray: errors }
+    : await calculateStats(donorRecord, existentDonor);
```
joneubank: this function name is confusing... its not actually calculating stats, just checking for new/updated/not-updated status

src/submission/validation-clinical/donor.ts
```diff
@@ -1,28 +1,119 @@
...
+      ),
+    );
+  }
+}
+
+// cases (similar for other clinical types with minor difference in case 1)
+// 1 not changing gender and new clinicalInfo <=> new
+// 2 changing specimenType or tnd or changing clinicalInfo <=> update
+// 3 not new or update <=> noUpdate
+async function calculateStats(
```
joneubank: lets rename to `checkForUpdates`

src/submission/validation-clinical/donor.ts
```diff
@@ -1,28 +1,119 @@
...
+  record: DeepReadonly<SubmittedClinicalRecord>,
+  donor: DeepReadonly<Donor>,
+): Promise<ValidatorResult> {
+  const clinicalInfo = donor.clinicalInfo;
+
+  // no updates to specimenType or tnd but there is now existent clinicalInfo, new
+  if (donor.gender === record[FieldsEnum.gender] && !clinicalInfo) {
```
joneubank: `{}` is truthy, hows about `_.isEmpty(clinicalInfo)`?

src/submission/validation-clinical/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+    index,
+    info: {
+      ...info,
+      donorSubmitterId: newRecord[FieldsEnum.submitter_donor_id],
+      value: newRecord[fieldName],
+    },
+  };
+};
+
+export const buildSubmisisonUpdate = (
```
joneubank: typo

src/submission/validation-clinical/sample.ts
```diff
@@ -1,6 +1,91 @@
...
+    };
+  }
+  const specimen = utils.getRegisteredSubEntityInCollection(
+    FieldsEnum.submitter_specimen_id,
+    sampleRecord,
+    existentDonor.specimens,
+  ) as Specimen;
+  if (!specimen) {
+    return {
+      type: ModificationType.ERRORSFOUND,
```
joneubank: We're building this object a lot... should there be some code to do this in a consistent way?

joneubank: some method like you already have the `utils.buildSubmissionError`...

src/submission/submission-service.ts
```diff
@@ -255,26 +260,45 @@ export namespace operations {
...
     const relevantDonorsMap = await getDonorsInProgram(filters);
-    const clinicalTypeErrors = await validateSubmissionData(newDonorDataMap, relevantDonorsMap);
+    const validateResult: ClinicalTypeValidateResult = await validateSubmissionData(
+      newDonorDataMap,
+      relevantDonorsMap,
+    );
 
-    // collect and update data errors and stats
+    // update data errors/updates and stats
     let inValid: boolean = false;
```
joneubank: inside Valid? or just invalid 😛

src/submission/submission-service.ts
```diff
@@ -230,19 +234,20 @@ export namespace operations {
         `No active submission found with programId: ${programId} & versionId: ${versionId}`,
       );
     }
-    const newActiveSubmission = _.cloneDeep(exsistingActiveSubmission) as ActiveClinicalSubmission;
     if (
       exsistingActiveSubmission.state === SUBMISSION_STATE.VALID ||
```
joneubank: shouldn't this if statement really just be `if (existingActiveSubmission.state !== SUBMISSION_STATE.OPEN)`?

joneubank:
Did we change the response object structure from validate service? if so... swagger update?

src/submission/submission-service.ts
jaserud: thtat doesn't take into account SUBMISSION_STATE.INVALID

joneubank: If I read this correctly, this is checking that validation has already been run so we can return the current activeSubmission instead of re-running validation. If that is correct, then when this is INVALID you can just return that, no need to validate again?

PR #159, opened 2019-09-24
Commit Clinical Sumbission

joneubank:
Service for committing active clinical submisison


src/resources/swagger.yaml
```diff
@@ -268,19 +294,11 @@ paths:
...
-          required: true
-          schema:
-            type: string
-        - name: versionId
-          description: uuid used to ensure correct version of submission is being updated
-          in: path
-          required: true
-          schema:
-            type: string
+        - '#/components/parameters/PathProgramId'
```
jaserud: I think this has to be like this `- $ref: '#/components/parameters/PathProgramId'`. Same for all the other ones...

src/resources/swagger.yaml
```diff
@@ -389,6 +440,18 @@ paths:
...
+  - name: Health
+    description: Service status monitoring
+  - name: Schemas
+    description: View and manage the Data Schema used by Argo-Clinical
+  - name: Registration
+    description: Registering new Donors, Specimens, and Samples
+  - name: Submission
+    description: Uploading, Validating, and Committing clinical data
+  - name: Clinical Data
+    description:
```
jaserud: does it let you do empty strings for description?

src/clinical/donor-repo.ts
```diff
@@ -22,12 +22,19 @@ export type FindByProgramAndSubmitterFilter = DeepReadonly<{
...
+  programId: string;
+  submitterIds: string[];
+}>;
 export interface DonorRepository {
   findByProgramId(programId: string): Promise<DeepReadonly<Donor[]>>;
   deleteByProgramId(programId: string): Promise<void>;
   findByProgramAndSubmitterId(
     filters: DeepReadonly<FindByProgramAndSubmitterFilter[]>,
   ): Promise<DeepReadonly<Donor[]> | undefined>;
+  findByProgramAndSubmitterIdList(
```
jaserud: duplicate functionality with findByProgramAndSubmitterId()

src/submission/merge-submission.ts
```diff
@@ -0,0 +1,105 @@
...
+  donors: readonly DeepReadonly<Donor>[],
+) => {
+  const updatedDonors = _.cloneDeep(donors as Donor[]);
+  for (const entityType in activeSubmission.clinicalEntities) {
+    const entityData = activeSubmission.clinicalEntities[entityType];
+
+    // Find the donor that matches each record, and update the entity within that donor
+    entityData.records.forEach(record => {
+      const donorId = record.submitter_donor_id;
+      const donor = _.find(updatedDonors, ['submitterId', donorId]);
```
jaserud: could use donorMap, might save some time?

src/submission/submission-entities.ts
```diff
@@ -88,6 +88,11 @@ export interface CreateRegistrationRecord {
   readonly sampleType: string;
 }
 
+export interface CommitClinicalSubmissionCommand {
```
jaserud: This could actually be used for validate and approve, we could name this something more general?

src/submission/submission-to-clinical.ts
```diff
@@ -5,17 +5,63 @@
...
+    throw new Errors.InvalidArgument(
+      'Version ID provided does not match the latest submission version for this program.',
+    );
+  } else if (activeSubmission.state !== SUBMISSION_STATE.VALID) {
+    // confirm that the state is VALID
+    throw new Errors.StateConflict(
+      'Active submission does not have state VALID and cannot be committed.',
+    );
+  } else {
+    // We Did It! We have a valid active submission to commit! Everyone cheers!
```
jaserud: :beer:

src/submission/submission-to-clinical.ts
```diff
@@ -5,17 +5,63 @@
...
+    // Get all the current donor documents
+    const donorDTOs = await getDonorDTOsForActiveSubmission(activeSubmission);
```
jaserud: like we talked about could use `getDonorsInProgram()` from submission-service

src/submission/merge-submission.ts
```diff
@@ -0,0 +1,105 @@
...
+import { update } from './schema-api';
+const L = loggerFor(__filename);
+
+type ClinicalEnitityRecord = DeepReadonly<
+  Readonly<{
+    [key: string]: string;
+  }>
+>;
+
+export const mergeActiveSubmissionWithDonors = async (
```
jaserud: could call this function with a DonorMap instead of a donor array

src/submission/merge-submission.ts
```diff
@@ -0,0 +1,105 @@
...
+  return updatedDonors;
+};
+
+const updateDonorFromSubmissionRecord = (donor: Donor, record: ClinicalEnitityRecord) => {
+  donor.gender = record.gender;
+  donor.clinicalInfo = _.omit(record, ['program_id', 'submitter_donor_id', 'gender']);
+};
+const updateSpecimenRecord = (donor: Donor, record: ClinicalEnitityRecord) => {
+  // Find specimen in donor
+  const specimen = _.find(donor.specimens, [
```
jaserud: what happens if the specimen doesn't exist?

test/integration/submission/submission.spec.ts
```diff
@@ -802,6 +799,108 @@ describe('Submission Api', () => {
     });
   });
 
+  describe('clinical-submission: commit', function() {
```
jaserud: should we group the http code checks in one place and keep clinical-submission stuff in another?

src/resources/swagger.yaml
joneubank: It doesn't break anyways... I didn't know what to write here, so this is just placeholder.

src/clinical/donor-repo.ts
joneubank: I'll combine these two

src/submission/submission-to-clinical.ts
joneubank: Not going to refactor into using a map, but I did use the existing donor-repo fetch method, removed the redundant implementation

src/submission/merge-submission.ts
joneubank: Do you see a need to refactor this way or is it a matter of preference?

I think, if activeSubmission merging is slow this could be a future optimization. would want to make a general Array->Map method to handle this type of situation.

jaserud: was just a suggestion, not critical to refactor.

src/submission/merge-submission.ts
joneubank: If the specimen is not there we should break but in a controlled way. I'll add some checking and throw some intentional errors.

joneubank: Same with specimen.

PR #163, opened 2019-09-26
Fix/state transtions

jaserud:
small fix for activesubmission state transitions..

### DIFF
--- a/src/submission/submission-to-clinical.ts
+++ b/src/submission/submission-to-clinical.ts
@@ -45,7 +45,31 @@ export const commitClinicalSubmission = async (command: Readonly<ActiveSubmissio
     );
   } else {
     // We Did It! We have a valid active submission to commit! Everyone cheers!
-    performCommitSubmission(activeSubmission);
+
+    let pendingApproval: boolean = false;
+    for (const clinicalType in activeSubmission.clinicalEntities) {
+      if (activeSubmission.clinicalEntities[clinicalType].stats.updated.length > 0) {
+        pendingApproval = true;
+        break;
+      }
+    }
+    // check if there are updates, if so, change state to SUBMISSION_STATE.PENDING_APPROVAL and save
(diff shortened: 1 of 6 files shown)


src/submission/submission-to-clinical.ts
```diff
@@ -45,7 +45,31 @@ export const commitClinicalSubmission = async (command: Readonly<ActiveSubmissio
...
+        break;
+      }
+    }
+    // check if there are updates, if so, change state to SUBMISSION_STATE.PENDING_APPROVAL and save
+    if (pendingApproval) {
+      const newActiveSubmission = _.cloneDeep(activeSubmission) as ActiveClinicalSubmission;
+      newActiveSubmission.state = SUBMISSION_STATE.PENDING_APPROVAL;
+
+      // insert into database
+      const updated = await submissionRepository.updateProgramWithVersion(
```
joneubank: Maybe we can refactor this to have a `updateVersion` method in the submisssion-repo

src/submission/submission-to-clinical.ts
```diff
@@ -45,7 +45,31 @@ export const commitClinicalSubmission = async (command: Readonly<ActiveSubmissio
...
+    let pendingApproval: boolean = false;
+    for (const clinicalType in activeSubmission.clinicalEntities) {
+      if (activeSubmission.clinicalEntities[clinicalType].stats.updated.length > 0) {
+        pendingApproval = true;
+        break;
+      }
+    }
+    // check if there are updates, if so, change state to SUBMISSION_STATE.PENDING_APPROVAL and save
+    if (pendingApproval) {
+      const newActiveSubmission = _.cloneDeep(activeSubmission) as ActiveClinicalSubmission;
```
joneubank: can we move the check for updates and the update state codes into separate methods so that there isnt so much inline code

PR #173, opened 2019-10-03
Feature/list all schema & createdAt

jaserud:
Endpoint to get list of all schemas in clinical.
Also added date to clinical entities
and some more goodies

### DIFF
--- a/src/lectern-client/schema-manager.ts
+++ b/src/lectern-client/schema-manager.ts
@@ -3,6 +3,7 @@ import { SchemasDictionary, DataRecord, SchemaProcessingResult } from './schema-
 import { schemaClient as schemaServiceAdapter } from './schema-rest-client';
 import { schemaRepo } from './schema-repo';
 import { loggerFor } from '../logger';
+import { FileType } from '../submission/submission-api';
 const L = loggerFor(__filename);
 
 let manager: SchemaManager;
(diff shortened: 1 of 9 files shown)


src/lectern-client/schema-manager.ts
```diff
@@ -18,6 +19,16 @@ class SchemaManager {
   getCurrent = (): SchemasDictionary => {
     return this.currentSchema;
   };
+
+  getSubSchemasList = (): string[] => {
+    const subSchemasList: string[] = [];
+    this.currentSchema.schemas.forEach(subSchema => {
```
andricDu: avoid creating and then mutating variables when you can. If `currentSchema.schemas` is already an array, then you can probably do something like:
```typescript
return this.currentSchema.schemas.filter(s => s.name !== FileType.REGISTRATION)
```

src/routes/schema.ts
```diff
@@ -6,6 +6,8 @@ import * as schemaApi from '../submission/schema-api';
 const router = express.Router();
 
 router.get('/', wrapAsync(schemaApi.get));
+// returns a list of all sub-schemas except sample_registration
+router.get('/list', wrapAsync(schemaApi.getAllSchema));
```
andricDu: maybe make it plural? `getAllSchemas`
Just a suggestion

src/lectern-client/schema-manager.ts
```diff
@@ -21,13 +21,9 @@ class SchemaManager {
...
-    const subSchemasList: string[] = [];
-    this.currentSchema.schemas.forEach(subSchema => {
-      if (subSchema.name !== FileType.REGISTRATION) {
-        subSchemasList.push(subSchema.name);
-      }
-    });
-    return subSchemasList;
+    return this.currentSchema.schemas
+      .filter(s => s.name !== FileType.REGISTRATION)
+      .map(s => s.name);
```
andricDu: :+1:

PR #175, opened 2019-10-08
Feature/zipped templates

jaserud:
Adds the option to download `all.zip` from /submission/schema/template/all


andricDu:
any thoughts about adding a unit test for this?

src/resources/swagger.yaml
```diff
@@ -65,6 +65,10 @@ paths:
               schema:
                 example: 'program_id\tsubmitter_donor_id\tgender\tsubmitter_specimen_id\tspecimen_type\ttumour_normal_designation\tsubmitter_sample_id\tsample_type'
                 type: string
+            application/zip:
```
joneubank: I think we need a seperate path for the 'all' version, since it has a different return type. This also calls clear attention to this endpoint instead of declaring the value in the description.

src/submission/schema-api.ts
```diff
@@ -48,9 +50,28 @@ export const getTemplate = async (req: Request, res: Response) => {
...
+export const getAllTemplates = async (req: Request, res: Response) => {
+  const schemasDictionary = manager.instance().getCurrent();
+  const zip = new AdmZip();
+  res
+    .status(200)
+    .contentType('application/zip')
+    .attachment('all.zip');
+
+  schemasDictionary.schemas
+    .filter(s => s.name !== FileType.REGISTRATION)
```
joneubank: 👍

src/resources/swagger.yaml
jaserud: :+1:

PR #187, opened 2019-10-18
Add Clear Active Submission endpoints

joneubank:
**Description of changes**

DELETE method request that accepts the filetype to clear or the value `all` for clearing everything.

Swagger docs updated and tests included.

Some changes were made to existing tests that were failing but catching the AssertionErrors so they appeared to be passing.

For tests, added a new util method to do a find call to mongo db so we can test that the db is being updated correctly, not just the API returns.

**Type of Change**

- [X] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing

### DIFF
--- a/src/submission/submission-service.ts
+++ b/src/submission/submission-service.ts
@@ -11,6 +11,7 @@ import {
   CreateRegistrationRecord,
   CreateRegistrationCommand,
   CreateRegistrationResult,
+  ClearSubmissionCommand,
   FieldsEnum,
   ClinicalSubmissionCommand,
   MultiClinicalSubmissionCommand,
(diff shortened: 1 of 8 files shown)


src/submission/submission-service.ts
```diff
@@ -167,6 +168,50 @@ export namespace operations {
...
+            updatedClinicalEntites[clinicalType] = {
+              ...activeSubmission.clinicalEntities[clinicalType],
+              ...emptyStats,
+            };
+          }
+        }
+      }
+      const newActiveSubmission: ActiveClinicalSubmission = {
+        programId: command.programId,
+        state: SUBMISSION_STATE.OPEN,
```
jaserud: what about clear on `VALID` or does clear always go to `OPEN`?

test/integration/submission/submission.spec.ts
```diff
@@ -32,8 +33,11 @@ import { donorDao } from '../../../src/clinical/donor-repo';
 import { Donor } from '../../../src/clinical/clinical-entities';
 import { ErrorCodes, FileType } from '../../../src/submission/submission-api';
 import * as manager from '../../../src/lectern-client/schema-manager';
+import { submissionRepository } from '../../../src/submission/submission-repo';
```
jaserud: thats not used anymore?? :mag:

test/integration/submission/submission.spec.ts
```diff
@@ -844,6 +848,151 @@ describe('Submission Api', () => {
...
+        })
+        .catch(err => chai.assert.fail(err));
+
+      try {
+        file = fs.readFileSync(__dirname + '/specimen.tsv');
+      } catch (err) {
+        return err;
+      }
+
+      await chai
```
jaserud: it's possible to attach multiple files to one request by multiple `.attach` so you don't have to do them separately, unless if there is a reason for it...

PR #189, opened 2019-10-18
fix bug that we skipped converting string to integer

blabadi:
in lectern there was a faulty if statement that prevented integer strings
from being converted to correct type

**Description of changes**

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature

- [x] Manual testing

### DIFF
--- a/src/lectern-client/schema-functions.ts
+++ b/src/lectern-client/schema-functions.ts
@@ -129,23 +129,24 @@ const convertFromRawStrings = (
       return undefined;
     }
 
+    const valueType = field.valueType;
+    let formattedFieldValue = record[field.name];
+    const rawValue = formattedFieldValue;
+
     // convert field to match corresponding enum from codelist, if possible
-    if (field.restrictions && field.restrictions.codeList) {
+    if (field.restrictions && field.restrictions.codeList && valueType === ValueType.STRING) {
       const formattedField = field.restrictions.codeList.find(
         e => e.toString().toLowerCase() === record[field.name].toString().toLowerCase(),
       );
       if (formattedField) {
-        mutableRecord[field.name] = formattedField;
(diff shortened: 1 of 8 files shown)


test/integration/submission/submission.spec.ts
```diff
@@ -773,7 +773,7 @@ describe('Submission Api', () => {
           }
         });
     });
-    it('should return with appropriate stats', async () => {
+    it.only('should return with appropriate stats', async () => {
```
blabadi: will remove

src/lectern-client/schema-functions.ts
```diff
@@ -125,15 +125,17 @@ const convertFromRawStrings = (
...
+
+    // convert field to match corresponding enum from codelist, if possible
+    if (field.restrictions && field.restrictions.codeList) {
+      const formattedField = field.restrictions.codeList.find(
+        e => e.toString().toLowerCase() === record[field.name].toString().toLowerCase(),
+      );
+      if (formattedField) {
+        mutableRecord[field.name] = formattedField;
       }
       return undefined;
```
jaserud: are `codeLists` always strings...

blabadi: they could be numbers as far as I remember in Lectern, I don't think we have an integer code list yet though, plus this is not a code I added now, this is the existing code to normalize the enum format.

jaserud: I was only asking because we return after replacing with enum format, so for enum numbers it will remain string. Yeah but you are right we don't have those so I guess it doesn't matter

jaserud: I was only asking about this since we return undefined after enum replace, so integer/number enum would remain string. But yea your right I gues it doesn't really matter.

Issue #194, opened 2019-10-22
🐛 Cant use dots. i want dots.


rosibaj:
## Describe the bug
I cannot register ids with a `.` included. This should be allowed according to the field regex. 

## Steps To Reproduce

1. Go to https://platform-ui.qa.argo.cancercollaboratory.org/submission/program/RAWR-CA/sample-registration
2. Click on Upload File and attempt to register this data
```
program_id	submitter_donor_id	gender	submitter_specimen_id	specimen_tissue_source	tumour_normal_designation	submitter_sample_id	sample_type
RAWR-CA	rawr-1	male	sample-1	Bone marrow 	Normal	sample-id-1	Amplified DNA
RAWR-CA	rawr-2	male	sample-2	Lymph node	Primary tumour	sample-id-2	ctDNA
RAWR-CA	rawr-3	female	sample-3	Bone marrow 	Normal	sample-id-3	Total DNA
RAWR-CA	rawr-4	Other	sample-4	Cerebrospinal fluid	Primary Tumour	sample-id-4	Total DNA
RAWR-CA	rawr-5	Other	sample-5	Cerebrospinal fluid	Primary Tumour	sample-id-5	Total DNA
RAWR-CA	rawr-5	Other	sample-5	Cerebrospinal fluid	Primary Tumour	sample-id-5.1	Total RNA
```
3. Nothing happens on the UI. Checking the response from endpoint, you see this error:
```
{
    "error": "InternalError",
    "message": "failed to save registration"
}
```
mongo is spitting this error. failed to save registrationkey rawr-5. must not contain .

## Expected behaviour
I can add dots to ids.

PR #215, opened 2019-10-28
Feature/194 submitter id dots

blabadi:
**Description of changes**
change of the registration data model, the submitter id was used as object key in the stats sub objects, and that caused mongo to complain since those can have dots in them

example
was:
```
stats: [
  newDonors: { 'abcd.1' : [1,2,3] }
] 
```
Now
```
stats: [
  newDonors: { submitterId: 'abcd.1', rowNumbers: [1,2,3] }
] 
```

**_I'll update the gateway to handle this change too in a PR there_**

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature

- [x] Manual testing

### DIFF
--- a/src/resources/swagger.yaml
+++ b/src/resources/swagger.yaml
@@ -261,7 +261,7 @@ paths:
           $ref: '#/components/responses/ForbiddenError'
         '500':
           $ref: '#/components/responses/ServerError'
-  /submission/program/{programId}/clinical/upload:
+  /submission/program/{programId}/clinical/:
     get:
       tags:
         - Submission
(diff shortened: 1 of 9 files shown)


src/resources/swagger.yaml
```diff
@@ -261,7 +261,7 @@ paths:
           $ref: '#/components/responses/ForbiddenError'
         '500':
           $ref: '#/components/responses/ServerError'
-  /submission/program/{programId}/clinical/upload:
+  /submission/program/{programId}/clinical/:
```
blabadi: this is not related to the issue, but the word 'upload' here is not needed

jaserud: fyi, you'll have to update this in gateway too...

blabadi:
resolves #194

src/resources/swagger.yaml
blabadi: yeah check here https://github.com/icgc-argo/platform-api/pull/98

PR #227, opened 2019-11-04
validate unique records for registration & submission

jaserud:
**Description of changes**

- added duplicateRecordCheck in registration validate
- added clinicalSubmission unique record check (using unique ID for each clinical type)
- updated tests accordingly

**Type of Change**

- [x] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing


src/submission/validation.ts
```diff
@@ -140,6 +146,54 @@ export const usingInvalidProgramId = (
...
+  useAllRecordValues?: boolean, // use all record properties so it behaves like duplicate check
+): SubmissionValidationError[] => {
+  const identifierToIndexMap: { [k: string]: number[] } = {};
+  const indexToErrorMap: { [index: number]: SubmissionValidationError } = {};
+  const uniqueIdName = ClinicalUniqueIndentifier[clinicalType];
+
+  newRecords.forEach((record: any, index) => {
+    const uniqueIdentiferValue = useAllRecordValues ? JSON.stringify(record) : record[uniqueIdName];
+
+    if (!identifierToIndexMap[uniqueIdentiferValue]) {
```
blabadi: a general comment instead of if/else use if / return and decompose big functions into smaller private functions to be easier to follow

jaserud: :+1:

src/submission/validation.ts
```diff
@@ -140,6 +146,68 @@ export const usingInvalidProgramId = (
   }
   return [];
 };
+
+export const checkUniqueRecords = (
+  clinicalType: ClinicalEntityType,
+  newRecords: DeepReadonly<DataRecord[] | CreateRegistrationRecord[]>,
+  useAllRecordValues?: boolean, // use all record properties so it behaves like duplicate check
```
blabadi: is this used in any case, I see you picked sample_submitter_id as unique identifier for registration ?

blabadi: nvm I see that above

src/submission/validation.ts
```diff
@@ -69,6 +72,9 @@ export const validateRegistrationData = async (
     }
   }
 
+  // catch any rows that are exactly the same
+  errors = errors.concat(checkUniqueRecords(ClinicalEntityType.REGISTRATION, newRecords, true));
```
blabadi: so I'm now a bit confused, it's smart to pick sample submitter id as unique Id for a record in registration, so why do we still need the `useAllRecordsValues` ?

test/unit/submission/validation.spec.ts
```diff
@@ -809,7 +809,7 @@ describe('data-validator', () => {
             donorSubmitterId: 'AB1',
             gender: 'Female',
             programId: 'PEME-CA',
-            sampleSubmitterId: 'AM1',
+            sampleSubmitterId: 'AM2',
```
jaserud: changed this `sampleSubitterId` to not conflict with other record, since this test is only checking `gender` mutation

src/submission/validation.ts
jaserud: yea I have removed sample_registration from this check, we can check it in `conflictingNewSample` and we can use `NEW_SAMPLE_ID_CONFLICT` (I checked it with Rosi)

PR #245, opened 2019-11-13
Feature/232 detect regex & friends

blabadi:
**Description of changes**

move schema manager outside lectern client
introduce the ability to fetch another schema
change revalidating the donor by using schema process function
fetch schema based on the donor last valid schema
account for script, required, req fields changes
store the reasons of invalidation 
 modifed merge submission to avoid omitting fields
moved merge-submission and submission-to-clinical to a folder
added auth_disabled for local dev flow
add cache to store breaking changes schema names per version pair

**THE CORE OF THIS DIFF is here: src/submission/schema-migration/migration-manager.ts**
**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [ ] Manual testing


src/submission/schema-migration/migration-manager.ts
```diff
@@ -83,54 +90,49 @@ const submitMigration = async (
...
+  }
+  const migrationId = migration._id;
+  const dryRun = migration.dryRun;
+
   while (!migrationDone) {
     let invalidCount = 0;
     let validCount = 0;
     const donors = await getNextUncheckedDonorDocumentsBatch(migrationId, 20);
-
+    const newSchemaVersion = newSchema.version;
```
jaserud: could probably set this once outside the loop, no...

src/submission/schema-migration/migration-manager.ts
```diff
@@ -172,80 +182,108 @@ const getNextUncheckedDonorDocumentsBatch = async (migrationId: string, limit: n
...
-      }
-    });
-
-    if (invalidFound) {
-      return true;
+  if (schemaName == 'specimen') {
+    const clinicalRecords = donor.specimens
+      .map(sp => {
+        if (sp.clinicalInfo) {
+          return prepareForSchemaReProcessing(sp.clinicalInfo, donor.submitterId);
```
jaserud: This won't prepare `submitter_specimen_id`, which is not saved during merge. I think it might be easier to save everything in record in DB.

src/submission/schema-migration/migration-manager.ts
```diff
@@ -172,80 +182,108 @@ const getNextUncheckedDonorDocumentsBatch = async (migrationId: string, limit: n
...
+
+  const schemaNamesWithBreakingChanges = _.uniqBy(
+    invalidatingFields.map((inf: any) => {
+      return inf.fieldPath.split('.')[0];
+    }),
+    (e: string) => e,
+  );
 
-  return flag;
+  for (const schemaName of schemaNamesWithBreakingChanges) {
```
jaserud: looks like we run the schema validation even if its just one breaking change, so wouldn't it be easier to just find only the `clinical_type` with breaking changes. do we need to find the specific fields here?? maybe in analysis we can keep track of schema names that are not valid?

src/submission/schema-migration/migration-manager.ts
blabadi: good catch will fix

src/submission/schema-migration/migration-manager.ts
blabadi: will remove

src/submission/schema-migration/migration-manager.ts
blabadi: I checked the reason, we need to analyze by field first to see if it required or not in some cases check line 386 for example in this file

jaserud: fair enough, and looks like `breakingChangesEntitesCache` is keeping track of `clinical_types` between version changes so :+1:

blabadi:
I need to resubmit this, found a a gap in the way diffs are returned and need modifications will update later today

PR #265, opened 2019-11-26
Feature/234 open subs

blabadi:
**Description of changes**
add check open submissions to schema migration
add state for invalid by migration.
add new method to migrate active submissions.
keep track of checked submissions
store invalid submissions ids

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [ ] Manual testing

### DIFF
--- a/src/submission/schema/schema-manager.ts
+++ b/src/submission/schema/schema-manager.ts
@@ -20,6 +20,7 @@ import {
   ClinicalEntityType,
   MultiClinicalSubmissionCommand as SubmitMultiClinicalCommand,
   NewClinicalEntity,
+  RevalidateClinicalSubmissionCommand,
 } from '../submission-entities';
 import { notEmpty, Errors } from '../../utils';
 import _ from 'lodash';
@@ -28,7 +29,7 @@ const L = loggerFor(__filename);
 let manager: SchemaManager;
 
 class SchemaManager {
-  private currentSchema: SchemasDictionary = {
+  private currentSchemaDictionary: SchemasDictionary = {
     schemas: [],
     name: '',
     version: '',
@@ -36,15 +37,15 @@ class SchemaManager {
   constructor(private schemaServiceUrl: string) {}
 
   getCurrent = (): SchemasDictionary => {
-    return this.currentSchema;
+    return this.currentSchemaDictionary;
   };
 
   getSubSchemasList = (): string[] => {
-    return this.currentSchema.schemas.map(s => s.name);
+    return this.currentSchemaDictionary.schemas.map(s => s.name);
   };
 
-  getSubSchemaFieldNamesWithPriority = (definition: string): FieldNamesByPriorityMap => {
-    return service.getSubSchemaFieldNamesWithPriority(this.currentSchema, definition);
+  getSchemaFieldNamesWithPriority = (definition: string): FieldNamesByPriorityMap => {
+    return service.getSchemaFieldNamesWithPriority(this.currentSchemaDictionary, definition);
   };
 
   /**
@@ -72,7 +73,7 @@ class SchemaManager {
   analyzeChanges = async (oldVersion: string, newVersion: string) => {
     const result = await changeAnalyzer.fetchDiffAndAnalyze(
       this.schemaServiceUrl,
-      this.currentSchema.name,
+      this.currentSchemaDictionary.name,
       oldVersion,
       newVersion,
     );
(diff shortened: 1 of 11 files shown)


src/submission/submission-service.ts
```diff
@@ -279,51 +293,162 @@ export namespace operations {
...
       batchErrors: [...dataToEntityMapErrors, ...fieldNameErrors],
     };
   };
 
+  /**
+   * checks existing submission records against a target schema
+   * @param command RevalidateClinicalSubmissionCommand
+   * @param targetSchema the schema that the submission records will be validated with.
+   */
+  export const revalidateClinicalSubmission = async (
```
blabadi: new method to revalidate submission

src/submission/submission-service.ts
```diff
@@ -279,51 +293,162 @@ export namespace operations {
...
+
+    const newActiveSubmission: ActiveClinicalSubmission = {
+      programId: command.programId,
+      state: state,
+      version: '', // version is irrelevant here, repo will set it
+      clinicalEntities: existingClinicalEntities,
+      updatedBy: exsistingActiveSubmission.updatedBy,
+    };
+
+    // insert into database
```
blabadi: bad comment will remove

src/submission/schema/schema-manager.ts
```diff
@@ -268,20 +282,77 @@ namespace MigrationManager {
...
+
+    for (const sub of submissions) {
+      if (
+        sub.state == SUBMISSION_STATE.INVALID ||
+        sub.state == SUBMISSION_STATE.INVALID_BY_MIGRATION
+      ) {
+        continue;
+      }
+
+      const checkedSubmission = migration.checkedSubmissions.find(x => {
```
joneubank: Unless I'm misreading, the method provided to find here never returns a value, no? just assigns the `sub._id` to `x._id` for every element in checkedSubmissions.

src/submission/schema/schema-manager.ts
```diff
@@ -268,20 +282,77 @@ namespace MigrationManager {
...
+      const result = await submissionService.operations.revalidateClinicalSubmission(
+        command,
+        newSchema,
+        dryRun,
+      );
+      if (!result.submission) {
+        continue;
+      }
+
+      if (result.submission.state == 'INVALID_BY_MIGRATION') {
```
joneubank: Should use `===` unless `==` is intentionally needed

src/submission/schema/schema-manager.ts
```diff
@@ -268,20 +282,77 @@ namespace MigrationManager {
...
+  const revalidateOpenSubmissionsWithNewSchema = async (
+    migration: DictionaryMigration,
+    newSchema: SchemasDictionary,
+    dryRun: boolean,
+  ) => {
+    const submissions = await submissionService.operations.findActiveClinicalSubmissions();
+
+    for (const sub of submissions) {
+      if (
+        sub.state == SUBMISSION_STATE.INVALID ||
```
joneubank: Should use `===` unless `==` is intentionally needed

src/submission/schema/schema-manager.ts
```diff
@@ -268,20 +282,77 @@ namespace MigrationManager {
...
+      }
+
+      if (result.submission.state == 'INVALID_BY_MIGRATION') {
+        migration.invalidSubmissions.push({
+          programId: sub.programId,
+          id: sub._id,
+        });
+      }
+
+      migration = await migrationRepo.update(migration);
```
joneubank: Are we modifying the migration variable that is a method argument? should we instead build this without side effects, returning a new migration object?

src/submission/submission-service.ts
```diff
@@ -279,51 +293,161 @@ export namespace operations {
...
     // insert into database
-    const updated = await submissionRepository.updateSubmissionWithVersion(
+    let updated = (await updateSubmissionWithVersionOrDeleteEmpty(
       command.programId,
       exsistingActiveSubmission.version,
       newActiveSubmission,
-    );
+    )) as ActiveClinicalSubmission;
+
+    if (updated == undefined) {
```
joneubank: This one needs to have `===`, pretty sure lots of things can match to undefined on a `==`

src/submission/schema/schema-manager.ts
blabadi: rookie mistake, should been ===

src/submission/schema/schema-manager.ts
blabadi: I need to update the migration document in every loop iteration, and this is a private method it just to break down the code .. im not sure if this is considered a side effect because most of our business logic interacts with a repository class of some kind.. I'm not sure that I understand the problem

src/submission/schema/schema-manager.ts
joneubank: Compact way to write this in the future:
`find(x => x._id === sub._id)`

src/submission/schema/schema-manager.ts
joneubank: Being private (used only locally here) means no risk of being called in the future and causing unexpected changes. Good enough for me.

PR #266, opened 2019-11-27
Feature/257 lock submission system

jaserud:
**Description of changes**
- set up persistedConfig api, service and repo
- added disable submisison to persisted configurations
- migration automatically sets/unsets `submissionDisabled`

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing
- [X] Regression tests completed and passing (double check number of tests).


src/resources/swagger.yaml
```diff
@@ -19,6 +19,48 @@ paths:
...
+          required: true
+          schema:
+            type: boolean
+      responses:
+        '500':
+          description: 'Failed to set submissionDisabled'
+        '200':
+          description: 'Sample registration and Clinical Submission disabled state has been set'
+  /submission/persistedConfig/replace:
+    post:
```
blabadi: I think it can be `patch` and `/submission/persisted-config`

src/resources/swagger.yaml
```diff
@@ -19,6 +19,48 @@ paths:
...
+        '200':
+          description: 'Sample registration and Clinical Submission system disabled state'
+    post:
+      tags:
+        - Persisted Configurations
+      summary: Disable sample registration and clinical submission
+      parameters:
+        - name: setDisabled
+          description: the boolean state to set for submissionDisabled
+          in: query
```
blabadi: usually post is to create new stuff, if you want to update it should be put / patch and the data should be in the body

src/submission/submission-api.ts
```diff
@@ -1,26 +1,25 @@
...
 import { HasFullWriteAccess, HasProgramWriteAccess } from '../auth-decorators';
-import jwt from 'jsonwebtoken';
 import _ from 'lodash';
 import { batchErrorMessage } from './submission-error-messages';
 const L = loggerFor(__filename);
 
 class SubmissionController {
   @HasProgramWriteAccess((req: Request) => req.params.programId)
   async getRegistrationByProgramId(req: Request, res: Response) {
+    if (await submissionSystemIsDisabled(res)) return;
```
blabadi: I'm not sure if the checks should be here

blabadi: I would rather put them in the submission service

jaserud: hmm, I thought about that too, but it has to be done in `submission-to-clinical` too because `commit` & `approve` are their. Also it didn't really make sense to do any processing if the system is disabled...

PR #281, opened 2019-11-29
add program id validation to clinical upload

blabadi:
stream records to process different validations
move stuff around

**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature

- [ ] Manual testing


src/submission/validation-clinical/validation.ts
```diff
@@ -0,0 +1,623 @@
+import { DonorMap, Specimen, Sample, Donor } from '../../clinical/clinical-entities';
```
blabadi: I moved this file from /submission to /validation-clinical .. no new content

src/submission/submission-service.ts
```diff
@@ -90,43 +93,44 @@ export namespace operations {
...
-    }
-
+    let validRecordsAccumulator: DeepReadonly<any[]> = [];
     // check the program id if it matches the authorized one
     // This check is used to validate the program Id along with the schema validations
     // to save extra round trips
-    let programIdErrors: DeepReadonly<SubmissionValidationError[]> = [];
     command.records.forEach((r, index) => {
-      const programIdError = dataValidator.usingInvalidProgramId(
+      const schemaResult = schemaManager
```
blabadi: instead of two iterations over data, do one iteration

src/lectern-client/schema-functions.ts
```diff
@@ -57,34 +57,72 @@ export const process = (
...
+    `done processing all rows, validationErrors: ${validationErrors.length}, validRecords: ${processedRecords.length}`,
+  );
+
+  return F({
+    validationErrors,
+    processedRecords,
+  });
+};
+
+export const process = (
```
blabadi: a new method to process one record instead of batch but keeps index information

jaserud: cool, but it looks a little weird, this function is processing one record, but it returns `prcessedRecords` which is an array that only contains the one record?? guess it makes it easy if both functions return arrays... Is that what you were going for?

src/submission/submission-service.ts
```diff
@@ -90,43 +93,44 @@ export namespace operations {
...
+      const programIdError = usingInvalidProgramId(
         ClinicalEntityType.REGISTRATION,
         index,
         r,
         command.programId,
       );
-      programIdErrors = programIdErrors.concat(programIdError);
+
+      unifiedSchemaErrors = unifiedSchemaErrors.concat(programIdError);
+      validRecordsAccumulator = validRecordsAccumulator.concat(schemaResult.processedRecords);
```
jaserud: same comment as earlier...

src/submission/submission-service.ts
```diff
@@ -504,6 +508,7 @@ export namespace operations {
     const validateResult: ClinicalTypeValidateResult = await validateSubmissionData(
       clinicalSubmissionRecords,
       relevantDonorsMap,
+      command.programId,
```
jaserud: why do we need to pass `command.programId` now, does it somehow use it??

src/lectern-client/schema-functions.ts
blabadi: I'll take a look .. I don't think I left it like that for a reason

src/submission/submission-service.ts
blabadi: Nope I think I added it here first before moving the programId validation logic to upload

PR #292, opened 2019-12-09
added therapy and treament to repo and donor

jaserud:
**Description of changes**
- setup treament/therapy in documents and donor
- made changes to keep tests passing

Take a look, let me know what you think.

**Type of Change**

- [ ] Bug
- [x] Refactor
- [x] New Feature

- [x] Manual testing


src/clinical/clinical-entities.ts
```diff
@@ -36,4 +34,18 @@ export interface Sample {
   submitterId: string;
 }
 
+export interface Treatment {
+  clinicalInfo?: ClinicalInfo;
+  chemotherapy: Array<Therapy>;
```
blabadi: the only difference between therapies is the name .. I'm thinking that we can generalize this .. a Treatment has a List of Therapies and each therapy has a name/type + clinical info

src/clinical/clinical-entities.ts
```diff
@@ -36,4 +34,18 @@ export interface Sample {
...
+  hormoneTherapy: Array<Therapy>;
+}
+
+export interface Therapy {
+  clinicalInfo?: ClinicalInfo;
+}
+
+export interface ClinicalInfo {
+  [field: string]: string | number;
```
blabadi: this could also be boolean | undefined

jaserud: (:+1: ) true, I just copy pasta'd what was already their for all clinicalInfo, will update..

blabadi: yes but you also generalized this, before it was specific to donor and specimen, which according to the schema didn't have boolean or other stuff

src/clinical/clinical-entities.ts
jaserud: hmm that's how I set it up at first, but then should `name/type` be constrain to an enum: ['hormoneTherapy', 'chemotherapy', ...] or is just `string` enough. I didn't want to define a SubClinicalEntity enum

src/clinical/clinical-entities.ts
jaserud: do we want to keep that information or is it ok to generalize it?

blabadi: it's ok

src/clinical/clinical-entities.ts
jaserud: ^ for that I mean in the mongoose document schema

blabadi: in mongoose  that can be a string

jaserud: cool, then I agree that makes more sense

PR #342, opened 2020-01-13
Feature/290 validation dependency

jaserud:
**Description of changes**

- updated all stub scehmas to latest due to this PR, tests using schema that are missing validation dependcies will/could fail.

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing

### DIFF
--- a/src/submission/schema/schema-manager.ts
+++ b/src/submission/schema/schema-manager.ts
@@ -5,13 +5,14 @@ import {
   SchemaProcessingResult,
   FieldNamesByPriorityMap,
   ChangeAnalysis,
+  SchemaDefinition,
 } from '../../lectern-client/schema-entities';
 import * as changeAnalyzer from '../../lectern-client/change-analyzer';
 import { schemaClient as schemaServiceAdapter } from '../../lectern-client/schema-rest-client';
 import { schemaRepo } from './schema-repo';
 import { loggerFor } from '../../logger';
 import { migrationRepo } from './migration-repo';
-import { DictionaryMigration } from './migration-entities';
+import { DictionaryMigration, ClinicalSchemaSpecs } from './migration-entities';
 import { Donor, ClinicalInfo } from '../../clinical/clinical-entities';
 import { DeepReadonly } from 'deep-freeze';
 import * as clinicalService from '../../clinical/clinical-service';
@@ -25,6 +26,7 @@ import {
 import { notEmpty, Errors, sleep } from '../../utils';
 import _ from 'lodash';
 import { getClinicalEntitiesFromDonorBySchemaName } from '../submission-to-clinical/submission-to-clinical';
+
 const L = loggerFor(__filename);
 
 let manager: SchemaManager;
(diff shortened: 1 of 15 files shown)


src/submission/validation-clinical/utils.ts
```diff
@@ -288,31 +291,17 @@ const getSubmissionErrorInfoObject = (
...
-
-  if (therapyType === ClinicalEntitySchemaNames.RADIATION) {
-    return (
-      treatmentType !== 'radiation therapy' &&
-      treatmentType !== 'combined chemo+radiation therapy' &&
-      treatmentType !== 'photodynamic therapy'
-    );
-  }
+  return !TreatmentTypeValuesMappedByTherapy[therapyType].find(
+    ttv => ttv.toString().toLowerCase() === treatmentType.toString().toLowerCase(),
```
blabadi: I'm not sure if we should lower case here, at this point it should've been normalized, did you encounter a case or think this is needed ?

src/submission/submission-entities.ts
```diff
@@ -365,3 +392,41 @@ export interface ClinicalSubmissionRecordsByDonorIdMap {
...
+  ],
+  [ClinicalEntitySchemaNames.RADIATION]: [
+    'Radiation therapy',
+    'Combined chemo+radiation therapy',
+    'Photodynamic therapy',
+  ],
+  [ClinicalEntitySchemaNames.HORMONE_THERAPY]: ['Hormonal therapy'],
+};
+
+export const DonorVitalStatusTimeSensitiveValues = ['Deceased'];
```
blabadi: I would put this as an enum / constant object and use the key in the code

src/submission/schema/schema-manager.ts
```diff
@@ -356,6 +365,85 @@ namespace MigrationManager {
     return closedMigration;
   };
 
+  const verifyNewSchemaIsValidWithDataValidation = async (
+    newSchemaDictionary: SchemasDictionary,
+  ): Promise<NewSchemaVerificationResult> => {
+    const verificationResult: NewSchemaVerificationResult = {};
+
+    Object.values(ClinicalEntitySchemaNames).forEach(clinicalEntityName => {
+      const clinicalEntityNewSchemaDef = newSchemaDictionary.schemas.find(
```
blabadi: what happens below if this is missing a whole schema, like donor

jaserud: if the a whole schema is missing then its schemaDefinition will be undefined and later on instead of getting all the fields in the schema it will be an empty array. so `checkClinicalEntityNewSchemaHasRequiredFields` will return all the fields in the enum we need for the missing entity in schema.

jaserud: same with missing field code list values

src/submission/submission-entities.ts
jaserud: I did thing of that but what if there are multiple values, e.g. `Dead` gets added so we are looking to match `Dead` or `Deceased`

src/submission/validation-clinical/utils.ts
jaserud: Yea ideally the record should match the codelist values by this point. I just did to `toLlowerCase()` to avoid potential bug. I'm not married to it I can remove it if you prefer.

PR #354, opened 2020-01-23
update notEmpty to consider empty object

jaserud:
**Description of changes**

- add empty object check in addition to null and undefined
- also updated migration test to check that there are no invalid donor errors

This bug never occurred before because migration schema validation only runs if there are breaking changes for a clinical entity schema, which never happened before.

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature

- [ ] Manual testing


jaserud:
Actually ignore for now.

PR #364, opened 2020-01-30
fix updates check

jaserud:
**Description of changes**

- fix clinical returning updates for files with no updates
- update tests

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature

- [ ] Manual testing

### DIFF
--- a/src/submission/validation-clinical/utils.ts
+++ b/src/submission/validation-clinical/utils.ts
@@ -13,13 +13,12 @@ import {
   ClinicalFields,
   TreatmentTypeValuesMappedByTherapy,
   ClinicalTherapyType,
-  DonorVitalStatusValues,
 } from '../submission-entities';
 import { DeepReadonly } from 'deep-freeze';
 import { validationErrorMessage } from '../submission-error-messages';
 import _ from 'lodash';
 import { DataRecord } from '../../lectern-client/schema-entities';
-import { Donor } from '../../clinical/clinical-entities';
+import { Donor, ClinicalInfo } from '../../clinical/clinical-entities';
 import { getSingleClinicalEntityFromDonorBySchemanName } from '../submission-to-clinical/submission-to-clinical';
 import { donorDao } from '../../clinical/donor-repo';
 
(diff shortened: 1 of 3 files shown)


src/submission/validation-clinical/utils.ts
```diff
@@ -112,14 +112,11 @@ const getSubmissionUpdates = (clinicalObject: any, record: SubmittedClinicalReco
     for (const fieldName in record) {
       // skip the index field
       if (fieldName == 'index') continue;
-      // new field added not in the old object
-      if (!clinicalObject[fieldName]) {
```
jaserud: fields values that were `null` in clinicalInfo were being returned as updates because this evaluated as true, which would've been fine if the record value wasn't also `null`

src/submission/validation-clinical/utils.ts
```diff
@@ -59,8 +59,8 @@ export const buildSubmissionUpdate = (
     index,
     info: {
       donorSubmitterId: newRecord[DonorFieldsEnum.submitter_donor_id],
-      newValue: `${newRecord[fieldName]}`, // we convert the value to string since lectern may converted it to non string (integer, boolean)
-      oldValue: `${oldValue}`,
```
jaserud: if the value is null this returned "null", not sure if thats what we wanted?

blabadi: I don't quite remember the order of how things changed, but this may have happened before the changes in lectern where I stopped it from populating default values to unsubmitted fields

src/submission/validation-clinical/utils.ts
blabadi: so my question is why would `record[fieldName]` be undefined/null ?

jaserud: I believe if you pass a header with no values in the column, that record[fieldName] will be undefined because the processed records from schema processing returns them like that. Then when undefined is stored in mongo it becomes null.

src/submission/validation-clinical/utils.ts
```diff
@@ -106,20 +105,20 @@ const checkForUpdates = (
...
     for (const fieldName in record) {
       // skip the index field
-      if (fieldName == 'index') continue;
-      // new field added not in the old object
-      if (!clinicalObject[fieldName]) {
-        submissionUpdates.push(buildSubmissionUpdate(record, '', fieldName));
-        continue;
-      }
-      // this is assuming that the field name record and clinicalInfo both have snake casing
+      if (fieldName == 'index' || (!clinicalObject[fieldName] && !record[fieldName])) continue;
```
blabadi: Can you also update the comment on why the other check is there

PR #368, opened 2020-02-05
Feat/358 calculate core fields

jaserud:
**Description of changes**

- each clinical entity stores their own stats in clincialInfoStats
- Donor stores aggregated stats of all submitted & expected fields
- `stat-calculator` subtracts original clincal entity stat values from aggregatedStats and adds the new ones
  - this occurs after merge step (which covers both commit & approve)

Note:these stats don't need to be calculated for sample-registration, even though there are core fiel

**Type of Change**

- [ ] Bug
- [x] Refactor
- [x] New Feature

- [x] Manual testing

### DIFF
--- a/src/clinical/clinical-entities.ts
+++ b/src/clinical/clinical-entities.ts
@@ -8,7 +8,7 @@ export interface Donor {
   specimens: Array<Specimen>;
   clinicalInfo?: ClinicalInfo;
   clinicalStats?: ClinicalStats;
-  primaryDiagnosis?: ClinicalObject | undefined;
+  primaryDiagnosis?: ClinicalEntity | undefined;
   followUps?: Array<FollowUp>;
   treatments?: Array<Treatment>;
   createdAt?: string;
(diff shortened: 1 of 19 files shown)


src/submission/submission-to-clinical/merge-submission.ts
```diff
@@ -1,17 +1,27 @@
 import { DeepReadonly } from 'deep-freeze';
-import { Donor, Treatment, ClinicalInfo, FollowUp } from '../../clinical/clinical-entities';
+import {
```
jaserud: it looks like a lot of refactoring in this file, but its mostly just moving things around to make things more consistent (and hopefully easier to follow). The main thing being added here is that each `update` function now returns the clinicalEntity it updated, so that we can update the stats in that clinicalEntity and the aggregatedStats in the donor.

src/clinical/clinical-entities.ts
```diff
@@ -7,11 +7,29 @@ export interface Donor {
...
   updatedAt?: string;
+  aggregatedInfoStats?: AggregateClinicalInfoStats;
+}
+
+export interface AggregateClinicalInfoStats {
+  submittedCoreFields: number;
+  submittedExtendedFields: number;
+  expectedCoreFields: number;
+  expectedExtendedFields: number;
+  // normalSamples: number;
```
blabadi: no need for commented code imo, we will remember to add those eventually

src/submission/submission-to-clinical/merge-submission.ts
blabadi: Refactoring is a good thing

PR #409, opened 2020-03-09
add new required fields, update sample schema

wajiha-oicr:
**Description of changes**

- Added two additional required fields for specimen
- Also updated sample-schema with v4 because v3 was getting outdated

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature

- [x] Manual testing


sampleFiles/sample-schema.json
```diff
@@ -3021,6 +3021,2196 @@
       ],
```
blabadi: is the schema used by the tests updated too?

wajiha-oicr: those were already updated in an earlier PR. changing the sample schema was more for making local testing easier (i.e when pointing to the file rather than setting lectern up locally)

blabadi: you can make this version 1.0, and remove the old stuff here. this file was just meant to hold 1 version anyway we can use lectern container if we need locally to get diffs

blabadi: @jaserud  are you ok with that? do you rely on this file as it is right now ?

blabadi: I've checked the test file it has the same thing, we are adding new versions that are not used by the test, currently tests look at version 1 and 2 only (2 is used for a migration test)  so basically 3, 4 are useless, if we want to update (which helps keeping our test up to date with new possible cases)  we should update 1 and 2 only, no need to maintain more than that

wajiha-oicr: Are you talking about the schema for the tests? Because it seems like v3 and v4 are used in some tests (https://github.com/icgc-argo/argo-clinical/blob/396_createUnitTest/test/integration/submission/submission.spec.ts#L2243 and https://github.com/icgc-argo/argo-clinical/blob/396_createUnitTest/test/integration/submission/submission.spec.ts#L2186).

With regards to the sample schema, I'm okay with just maintaining one v1 but would need to see what @jaserud thinks.

blabadi: oh I see, ok then, It's out of the scope of this PR to fix what is existing in the tests. 
let's check when Jaser is back regarding the sample Schema

jaserud: @wajiha-oicr I don't have any problems with just holding one schema in sample-schema. The ones in the test however are being used so might be easier to leave them (I think there is a ticket to clean up tests maybe it could be done then). I think I remember doing it that way so any changes to the schema 1.0 wouldn't affect the other tests.

wajiha-oicr: Thanks Jaser. So I'll just update sample_schema with the latest dictionary and call it v1, deleting the rest.

wajiha-oicr:
In https://github.com/icgc-argo/argo-clinical/pull/409/commits/8f3b7bf070014b4fde683c77d77f029fda5f90f6 I edited the sample schema to only hold one dictionary version 1.0, which is the latest [28.0] version from lectern.

https://github.com/icgc-argo/argo-clinical/blob/8f3b7bf070014b4fde683c77d77f029fda5f90f6/sampleFiles/sample-schema.json

Tested a few endpoints with postman and things seem to be working as intended.

PR #415, opened 2020-03-10
✨ adds huge sample registration

hlminh2000:


wajiha-oicr:
@jaserud you have to check each line and make sure it's all correct

jaserud:
@wajiha-oicr already on it, I'm on line 186, only 20167 to go

PR #419, opened 2020-03-11
added topic creation

jaserud:
**Description of changes**

- added topic creation before producing messages
- added env variables for partition count and replication

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing


.env.example
```diff
@@ -52,4 +52,7 @@ KAFKA_MESSAGING_ENABLED=false
 # if messaging is set to false, values below don't need to be set
 KAFKA_CLIENT_ID=
 KAFKA_BROKERS=broker:1,broker:2,broker:3
-KAFKA_TOPIC_PROGRAM_UPDATE=
\ No newline at end of file
+KAFKA_TOPIC_PROGRAM_UPDATE=
+KAFKA_TOPIC_PROGRAM_UPDATE_CONFIG_PARTITIONS=1
+# number of replications can't be greater then available brokers
+KAFKA_TOPIC_PROGRAM_UPDATE_CONFIG_REPLICATIONS=2
```
blabadi: shouldn't this be 3 ?

src/submission/submission-updates-messenger.ts
```diff
@@ -21,48 +21,63 @@ class DummyMessenger implements SubmissionUpdatesMessenger {
...
+    this.kafka = new Kafka({ clientId, brokers });
+    this.programUpdateTopic = topics;
   }
 
   sendProgramUpdatedMessage = async (programId: string) => {
     L.info('KafkaMessenger called to send message to broker for updated program: ' + programId);
-    await this.producer.connect().catch(this.producerErrorHandler);
-    await this.producer
+
+    await this.createMissingProgramUpdateTopic().catch(this.errorHandler);
```
blabadi: shouldn't this be part of the bootstrap ?

jaserud: but, is it a one time thing though what if the topic doesn't exist when clinical tries to produce?

hlminh2000: why don't we just do this in the constructor? the topic just needs to be created once, it just adds extra overhead if we do this for every event

blabadi: it is a one time thing, like databases, why would we assume that topics would disappear? did you notice that while testing ?

hlminh2000: also once the topic is created, admin connection should be closed.

jaserud: yea that makes sense will change :+1:

jaserud: yea Bashar said the same thing, it makes sense to do it that way will change :+1:

src/submission/submission-updates-messenger.ts
```diff
@@ -20,49 +20,62 @@ class DummyMessenger implements SubmissionUpdatesMessenger {
   };
 }
 
-class KafkaMessenger implements SubmissionUpdatesMessenger {
-  private producer: Producer;
-  private knownTopics: ExpectedTopicsMap;
+const createKafkaMessenger = async (
```
jaserud: I've changed this into a function that returns the object, because I wanted it to run the create topic during messenger object creation and awaits can't be used in constructors

hlminh2000: Welcome to the dark side :)))

jaserud: Its a good thing we have darkvision :wink:

PR #433, opened 2020-03-20
send messages from migration program updates

jaserud:
**Description of changes**

- Added array of programs being updated to migration object
- Messages for each program in array is sent after donors are migrated
- Added a spy in migration tests on the send function to verify it is called as expected

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing

### DIFF
--- a/src/submission/schema/schema-manager.ts
+++ b/src/submission/schema/schema-manager.ts
@@ -621,7 +621,7 @@ namespace MigrationManager {
   ) => {
     const programIsBeingUpdated =
       donorBeforeMigration.schemaMetadata.isValid !== donorAfterMigration.schemaMetadata.isValid ||
-      !_.isEqual(donorBeforeMigration, donorAfterMigration);
+      !_.isEqual(donorBeforeMigration.aggregatedInfoStats, donorAfterMigration.aggregatedInfoStats);
     if (programIsBeingUpdated) {
       programsWithChanges.add(donorAfterMigration.programId);
     }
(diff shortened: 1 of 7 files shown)


src/submission/schema/schema-manager.ts
```diff
@@ -361,7 +363,11 @@ namespace MigrationManager {
 
     const migrationAfterDonorCheck = await checkDonorDocuments(migration, newTargetSchema);
 
-    const migrationAfterSubmissionsCheck = await revalidateOpenSubmissionsWithNewSchema(
+    migrationAfterDonorCheck.programsWithDonorUpdates?.forEach(program =>
+      messenger.getInstace().sendProgramUpdatedMessage(program),
```
blabadi: this doesn't throw errors right ?

jaserud: Hmm the only one I can think of is if the instance is undefiend which it shouldn't be. All the asyn functions called inside of `sendMessage` are handled with `.catch`. But I guess better safe then sorry. so I could add catch logic here. Also my bad typo should be `getInstance`, will fix.

blabadi: note that when you call for each for something that returns a promise, it will not await on it

jaserud: I don't think we need to wait on the messages being sent though right? The only reason to wait would be if we wanted to know whether the message reached kafka, but I created a seperate ticket to handle errors if messages don't reach kafka.

blabadi: I think so too, should be safe now after the try catch

PR #437, opened 2020-03-25
More migration unit tests

wajiha-oicr:
**Description of changes**
- Added migration unit tests to test behaviour for various kinds of (potentially breaking) migration changes
- tests were made from specs listed [here](https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Submission+System+Technical+Architecture) 
- Dynamically generated a new schema for each change for independent testing
    - the changes to the schemas are clearer to understand
    - the schemas don't contain various types of changes bundled together, making them easier to unit test
    - kept three legacy schemas, appended the dynamic schemas after to make a custom migration stub-schema

- Not every behaviour expectation for the migration scenarios has clear specs or implementation, so tests are subject to change

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [x] Tests

- [x] Manual testing


test/integration/submission/submission-schema-migration.spec.ts
```diff
@@ -229,85 +244,293 @@ describe('schema migration api', () => {
...
+      });
+
+      const res = await getAllMigrationDocs();
+      const donors = await findInDb(dburl, 'donors', {});
+
+      const [migration] = res.body;
+      migration.toVersion.should.equal('8.0');
+
+      migration.stage.should.not.equal('FAILED');
+      migration.should.not.have.property('newSchemaErrors');
```
jaserud: might want to make this a const, its used a lot in this file. also isn't their a const for `FAILED`

test/integration/submission/submission-schema-migration.spec.ts
```diff
@@ -229,85 +244,293 @@ describe('schema migration api', () => {
...
+        // migration will fail
+        schema[0].version.should.eq(startingSchemaVersion);
+      });
+
+      const res = await getAllMigrationDocs();
+      const [migration] = res.body;
+      migration.toVersion.should.equal('9.0');
+
+      migration.stage.should.equal('FAILED');
+      migration.should.have.property('newSchemaErrors');
```
jaserud: same here as before

test/integration/submission/submission-schema-migration.spec.ts
```diff
@@ -229,85 +244,293 @@ describe('schema migration api', () => {
...
+            { fieldName: DonorFieldsEnum.vital_status, missingCodeListValues: ['Deceased'] },
+          ],
+        },
+        [ClinicalEntitySchemaNames.REGISTRATION]: {
+          missingFields: [SampleRegistrationFieldsEnum.specimen_type],
+          invalidFieldCodeLists: [],
+        },
+      });
+    });
+    it.skip('should reject the new schema where the field type was changed', async () => {
```
jaserud: :eyes:, hmm this would only apply to fields that are used in validation eh. It sounds to me like the logic for checking if the schema should be rejected needs to consider the diff and not just the list of fields clinical is aware of... since this test is being skipped maybe make a ticket so it gets done (if not made already)

test/integration/submission/submission-schema-migration.spec.ts
```diff
@@ -23,10 +34,14 @@ chai.use(require('chai-http'));
 chai.use(require('deep-equal-in-any-order'));
 chai.should();
 
+// legacy field name
+const TUMOUR_STAGING_SYSTEM = 'tumour_staging_system';
 const schemaName = 'ARGO Clinical Submission';
 const startingSchemaVersion = '1.0';
 
 describe('schema migration api', () => {
+  buildDynamicStubSchema();
```
jaserud: I don't know this feels weird, do we need to build the migration stub schema every time the test is run. The dynamic schema building is cool since the stub schemas have to be updated with certain code changes... but it seems to me like this is a one time thing, no? Or is this something we want to do everytime.

test/integration/submission/migration_utils/schema_builder.ts
```diff
@@ -0,0 +1,142 @@
...
+export const buildDynamicStubSchema = () => {
+  const convertToDictionaryNotation = (schemaVersion: Object) => {
+    return Object.values(schemaVersion).map(schema => {
+      const { fields, ...rest } = schema;
+      const newFields = Object.values(schema.fields);
+      return _.assign(rest, { fields: newFields });
+    });
+  };
+
+  const legacyDictionaries = [
```
jaserud: It sounds to me like these are the same stubs in the schema-stub file, if so why not just import that??

test/integration/submission/migration_utils/schema_builder.ts
```diff
@@ -0,0 +1,142 @@
...
+    },
+
+    {
+      schemas: schemaV3,
+      name: DICTIONARY_NAME,
+      version: '3.0',
+    },
+  ];
+
+  // Change 8.1.1.1 adding new enum value Adding an enum to the codelist
```
jaserud: lol I don't understand the numbering `8.x.x.x`, is this the wajiha decimal system? or am I missing something else.

test/integration/submission/submission-schema-migration.spec.ts
```diff
@@ -229,85 +244,293 @@ describe('schema migration api', () => {
...
-        missingFields: [SampleRegistrationFieldsEnum.specimen_type],
-        invalidFieldCodeLists: [],
-      },
+      const res = await getAllMigrationDocs();
+      const [migration] = res.body;
+      migration.toVersion.should.equal('4.0');
+      migration.state.should.not.equal(migratationFailStage);
+      migration.should.not.have.property('newSchemaErrors');
+      migration.invalidDonorsErrors.should.be.empty;
+      migration.invalidSubmissions.should.be.empty;
```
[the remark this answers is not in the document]
wajiha-oicr: Yeah, that's a good point! Will try to commonize this to reduce the repetition.

test/integration/submission/submission-schema-migration.spec.ts
wajiha-oicr: good catch! I declared a const for it but I missed these!

test/integration/submission/migration_utils/schema_builder.ts
jaserud: oh I see it these the are numbers from the spec, nvm

test/integration/submission/submission-schema-migration.spec.ts
wajiha-oicr: I don't think there is a ticket for this, but you are correct, there probably should be (it was outlined in https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Submission+System+Technical+Architecture). I'll make the ticket.

test/integration/submission/submission-schema-migration.spec.ts
wajiha-oicr: Hm, I'm not sure. On one hand, not building it every time saves on the unnecessary repeated computation. But the potential issue with not calling on this function every time is that if somebody makes changes to the `schema-builder.ts` (say to add another schema), then they also have to remember to actually run the script so the json file will be rewritten to. If they forget then things relying on the latest schema won't work, which they may not realize?

Maybe I could put a comment? Or do something else?

test/integration/submission/migration_utils/schema_builder.ts
wajiha-oicr: yeah that make sense. I think I had initially planned to do something with these and then decided not to, then got stuck in tunnel vision. thanks!

test/integration/submission/migration_utils/schema_builder.ts
wajiha-oicr: come on you've never heard of the wajiha decimal system? 😡

test/integration/submission/submission-schema-migration.spec.ts
jaserud: But this process isn't so straight forward eh. It looks like you have to get the diffs from lectern and put them in stub-diffs.ts. So really you need to script/manually generate the stub-schemas anyways. There's a lot of places to forget to do something :stuck_out_tongue:.

blabadi:
I'm looking into this PR now, will take me some time

wajiha-oicr:
> I'm looking into this PR now, will take me some time

okay, but am currently adressing all of Jaser's feedback. I will push a commit again soon so maybe wait until then?

test/integration/submission/migration_utils/static-schema-data.ts
```diff
@@ -0,0 +1,4566 @@
+// Legacy, fixed Schema Versions from the stub-schema
+
+// The first version of the schema, but using object notation for schemas and fields for easy modification
+export const schemaV1_accessible = {
```
blabadi: what does `accessible` mean here?

wajiha-oicr: it uses object notation, where it uses keys for the file names and the fields. This makes it easier to extend upon the schema when trying to make a new version to migrate to, so you can easily modify the properties you want to change.

 Original schemas have them in arrays, so that would require knowledge of the array index

blabadi: did you build this manually ?

wajiha-oicr: I used a local script to generate it. I didn't keep the code but it used a lodash function.

blabadi: why not keep that script and generate on the fly ?

wajiha-oicr: I believe I had initially tried that but typescript produced errors when I tried to access the keys of the object since there was no type/interface defined for it

blabadi: you can use `any`

blabadi: but we already have alot of Types already defined in clinical service too that can be used

blabadi: this can be harder to maintain on the long run, that's why I'm asking about this

wajiha-oicr: I used `any` but then lost intellisense on the suggestion property and field names. I assumed version_1 would remain static but you are correct that it should be easily extendible. I think the best thing for me to do then is to fully define the type.

blabadi: I can help you in this tomorrow to figure out the best way without alot of manual typing just remind me if I forget please

wajiha-oicr:
the unit test being skipped doesnt have an implementation yet. ticket made here https://github.com/icgc-argo/argo-clinical/issues/446

README.md
```diff
@@ -30,6 +30,14 @@ We use a tool called migrate-mongo: https://www.npmjs.com/package/migrate-mongo
   - With this command, migrate-mongo will revert (only) the last applied migration
 - status: `npx migrate-mongo status`
 
+## Extending stub-schema for running DB migration Tests
+
+- extend upon the [schema builder](/test/integration/submission/migration_utils/schema_builder.ts) with the new schema
+- run the script using `npx ts-node schema_builder.ts` to generate the [migration-stub-schema.json](/test/integration/submission/migration-stub-schema.json)
```
blabadi: why is this not all done in memory ?

blabadi: this can be done before all the tests? isnt that how you had it before ?

wajiha-oicr: I changed it because it's still not purely automatic anyway; i.e. after making the schema, the diffs part of it needs to be brought in from lectern's diff calculator, and added to the file. so it's not reliable to be generated on the fly and we thought it would be better to just avoid the expensive computation before every test.

blabadi: it's fair that it's not automatic, yeah the diffs part is brutal and can be deceiving if lectern changed it's diff outputs somehow, our tests will pass when they shouldn't

blabadi: I'll create a ticket to run tests against a real lectern container

PR #456, opened 2020-04-02
Rewrite the icgc sample import

blabadi:
use 1 file export from song to import icgc samples
move sample files in a directory
update swagger
add 1 sanity test for import pre process

**Description of changes**

**Type of Change**
- [x] New Feature
- [x] Refactor

- [ ] Manual testing


test/integration/submission/submission.spec.ts
```diff
@@ -755,6 +755,36 @@ describe('Submission Api', () => {
...
+    this.beforeEach(async () => {
+      await clearCollections(dburl, ['donors', 'counters']);
+    });
+
+    it.only('should import legacy samples file', async () => {
+      let file: Buffer;
+      let rows: any[];
+      try {
+        file = fs.readFileSync(stubFilesDir + `/paca.icgc.samples.tsv`);
+        rows = (await TsvUtils.tsvToJson(stubFilesDir + `/paca.icgc.samples.tsv`)) as any;
```
jaserud, line 768: this test seems incomplete :stuck_out_tongue:. Shouldn't it also be calling the `/submission/icgc-import/` endpoint with the returned rows from the preprocess endpoint. Or was it just testing the precrocess returns 200...

src/submission/submission-service.ts
```diff
@@ -1005,17 +1006,22 @@ export namespace operations {
...
+    legacySamples: Readonly<LegacyICGCImportRecord>[],
+    programId: string,
+  ) {
     const result: Donor[] = [];
-
-    clinicalData.donors.forEach((d: { [k: string]: string }, i: number) => {
-      validateRequiredColumns(i, d, ['icgc_donor_id', 'submitted_donor_id']);
+    legacySamples.forEach((sampleRow: LegacyICGCImportRecord, i: number) => {
+      validateRequiredColumns(i, sampleRow, ['icgc_donor_id', 'submitted_donor_id', 'donor_sex']);
       result.push({
```
jaserud, line 1016: Isn't it possible to have duplicate donors in this result?? I see there are multiple rows in the file with same donorIds

jaserud: e.g. `DO35132`

test/integration/submission/submission.spec.ts
blabadi: yeah this test just calls preprocess

src/submission/submission-service.ts
blabadi: I'll check it

jaserud: It's not critical. It just the response JSON will fail on the import endpoint if there are duplicates.

blabadi: yes there will be duplicates, I didn't test the import manually, I submitted the  PR first :P

test/integration/submission/submission.spec.ts
```diff
@@ -755,6 +755,91 @@ describe('Submission Api', () => {
...
+        // 38 unique specimens
+        file = fs.readFileSync(stubFilesDir + `/paca.icgc.samples.tsv`);
+        rows = (await TsvUtils.tsvToJson(stubFilesDir + `/paca.icgc.samples.tsv`)) as any;
+      } catch (err) {
+        return err;
+      }
+
+      const response = await chai
+        .request(app)
+        .post('/submission/icgc-import/preprocess/ABCD-EF')
```
jaserud, line 778: didn't notice before but we don't verify the program name during import eh "ABCD-EF" !== "PACA-CA" :stuck_out_tongue: . But I guess that makes sense if a program name is different in argo then in icgc. Maybe a word of caution in swagger when using this endpoint or maybe it should accept the current program name and the program name to use...??

jaserud: This is not critical just something that could happen.

blabadi: so in this pre process endpoint they specify the new program name or fill in the same name.

then in the import they can import whatever they want, we don't need to verify anything because this completely new data. the pre process and import are stateless

jaserud: :+1: True, this endpoint is admins only so I don't think this is troubling.

PR #458, opened 2020-04-05
Optimize commit sample registration endpoint

blabadi:
**Description of changes**

Refactored the commit endpoint to be more performant
baseline : 3k document took 27-28 seconds on localhost 
now it takes between 9-12 secs on localhost

there is a bottleneck that I'm suspecting related to generating sequential unique id that even with parallelization of update/create statements for documents. but for this sprint this is a good improvement to begin with.

summary of all changes:
- use winston profiler for time logging.
- optimize getting related donors from db in one shot instead of per donor query.
- optimize saving documents in parallel and batches instead of awaiting each document.
- refactor performance test to cover different scenarios in one test case.

**Type of Change**
- [x] Refactor

- [x] Manual testing


src/utils.ts
```diff
@@ -163,6 +163,14 @@ export function mergeAndDeleteRemoved(obj: { [k: string]: any }, override: { [k:
   return;
 }
 
+export async function timeit<T>(tag: string, fn: () => T) {
```
blabadi, line 166: remove this

test/performance-test/submission.spec.ts
```diff
@@ -178,14 +179,14 @@ describe('Submission Api', () => {
...
   });
 });
 
-const timeit = async (fn: Function) => {
-  const start = new Date().getTime();
-  await fn();
-  const diff = (new Date().getTime() - start) / 1000.0;
-  console.log(diff);
-};
+// const timeit = async (fn: Function) => {
```
blabadi, line 187: remove this

test/performance-test/submission.spec.ts
```diff
@@ -159,15 +160,15 @@ describe('Submission Api', () => {
...
     it('Submit 3000 NEW samples into empty db & program', async function() {
-      await timeit(register3k);
+      const result = await timeit('register3k', register3k);
     });
 
     /**
      * Times committing 3000 donors registration into empty program
      */
-    it('Commit 3000 donors Registration into empty db', async function() {
+    it.only('Commit 3000 donors Registration into empty db', async function() {
```
blabadi, line 169: remove `.only`

src/utils.ts
jaserud: missed this :stuck_out_tongue:

blabadi:
> This looks alright to me. But it does look like the commit registration tests fail since the commit order of donors can be different then expected??

yep I'll have to compare out of order

blabadi:
passed and failed again.. will check

blabadi:
Should be ready now, i added chai exclude to avoid alot of lodash omitting

PR #471, opened 2020-04-20
Feat/465 update core calculation

jaserud:
**Description of changes**

- updated Core calculation in stat-calculator to new rules: https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Program+Dashboard+Page+Specs#ProgramDashboardPageSpecs-DataCalculations
- updated donor doc to better fit new core calculation
   - removed clinicalInfoStats since its not being used anymore (migration script)
- added new endpoint to allow manual override of core entities
- updated tests as needed

**Type of Change**

- [ ] Bug
- [x] Refactor
- [x] New Feature

- [x] Manual testing


src/submission/submission-to-clinical/stat-calculator.ts
```diff
@@ -1,139 +1,197 @@
...
-  ClinicalInfoStats,
-  AggregateClinicalInfoStats,
-} from '../../../src/clinical/clinical-entities';
-import { ClinicalEntitySchemaNames } from '../submission-entities';
-import { isNotAbsent, isEmpty, notEmpty } from '../../../src/utils';
-
-import * as schemaManager from '../schema/schema-manager';
-import { getClinicalObjectsFromDonor } from './submission-to-clinical';
-import { SchemasDictionary } from '../../../src/lectern-client/schema-entities';
+import { Donor } from '../../../src/clinical/clinical-entities';
```
jaserud, line 1: This file is really the bulk of the changes. I don't think its very clear to understand by looking at the diffs, so it might be easier to just see this file on its own.

src/resources/swagger.yaml
```diff
@@ -19,6 +19,49 @@ paths:
               schema:
                 type: object
 
+  /submission/admin-operations/recalculate-stats/{programId}/{submittedDonorId}:
```
blabadi, line 22: this should not be under submission, donors belong to /clinical/..
you can use `PATCH /clinical/donor/{donorId}/core-stats`
why are you using the submitter donor ID ?

blabadi: no need to mention `admin` in the url, it's validated in the authorization, I'd say it's better to keep roles names out of the api urls because something/someone else may use them in future

blabadi: you can however keep the tag as admin operations in swagger

src/submission/submission-api.ts
```diff
@@ -259,6 +259,26 @@ class SubmissionController {
     const donorsJson = await fsPromises.readFile(req.file.path, 'utf-8');
     return res.status(201).send(await submission.operations.adminAddDonors(JSON.parse(donorsJson)));
   }
+
+  // @HasFullWriteAccess()
```
blabadi, line 263: missed this ?

migrations/20200420215640-remove-clinical-info-stats.js
```diff
@@ -0,0 +1,19 @@
+export async function up(db) {
```
blabadi, line 1: the dbs got wiped .. it's good that you prepared this but I don't think it's is necessary now

sampleFiles/clinical/chemotherapy.tsv
```diff
@@ -1,4 +1,4 @@
 program_id	submitter_donor_id	submitter_treatment_id	chemotherapy_drug_name	chemotherapy_dosage_units	cumulative_drug_dosage
-TEST-CA	ICGC_0002	T_02	Still need it	mg/m²	4
-TEST-CA	ICGC_0003	T_03	Need list	µg/m²	20
-TEST-CA	ICGC_0001.1	T_04	Need list	µg/m²	35
\ No newline at end of file
+TEST-CA	ICGC_0002	T_02	Allopurinol	mg/m²	4
```
blabadi, line 2: does this work ?

src/app.ts
```diff
@@ -63,6 +64,7 @@ app.use(
 );
 
 /** Attach Routers */
+app.use('/submission/admin-operations', adminOp);
```
blabadi, line 67: as mentioned use the /clinical route

src/submission/submission-to-clinical/stat-calculator.ts
```diff
@@ -1,139 +1,197 @@
...
+      donor.specimens.some(sp => sp.tumourNormalDesignation === 'Normal') &&
+      donor.specimens.some(sp => sp.tumourNormalDesignation === 'Tumour');
+
+    coreStats[clinicalType] = tumorAndNormalExists
+      ? donor.specimens.map(sp => sp.clinicalInfo).filter(notEmpty).length / donor.specimens.length
+      : 0;
+  } else {
+    // for others we just need to find one clinical info for core entity
+    coreStats[clinicalType] =
+      getClinicalEntitiesFromDonorBySchemaName(donor, clinicalType).length >= 1 ? 1 : 0;
```
blabadi, line 48: why aren't we using true / false ?

src/clinical/clinical-entities.ts
```diff
@@ -19,17 +19,8 @@ export interface Donor {
...
-  expectedCoreFields: number;
-  expectedExtendedFields?: number;
-}
-
-export interface ClinicalInfoStats {
-  submittedCoreFields: number;
-  submittedExtendedFields?: number;
-  expectedCoreFields: number;
-  expectedExtendedFields?: number;
+  coreEntitiesStats: Record<CoreClinicalEntites, number>;
```
blabadi, line 22: not to repeat but I noticed  you only filling this with 0/1

sampleFiles/clinical/specimen.tsv
```diff
@@ -1,3 +1,2 @@
 program_id	submitter_donor_id	submitter_specimen_id	specimen_acquisition_interval	specimen_anatomic_location	central_pathology_confirmed	tumour_histological_type	tumour_grading_system	tumour_grading_system	tumour_grade	pathological_tumour_staging_system	pathological_stage_group	percent_tumour_cells	percent_proliferating_cells	percent_inflammatory_tissue	percent_stromal_cells	percent_necrosis
-TEST-CA	ICGC_0001.1	sub-sp-pacaau-123	200	Other	No	M-1111/22	Default	Default	aStringValue	Murphy	aStringValue	0.35	0.5	0.6	0.65	0.65
```
blabadi, line 2: was this deleted for testing?

src/clinical/clinical-entities.ts
```diff
@@ -1,4 +1,5 @@
 import { DeepReadonly } from 'deep-freeze';
+import { CoreClinicalEntites } from '../submission/submission-entities';
```
blabadi, line 2: clinical entities should not reference submission entities, this `CoreClinicalEntites` a clinical property not submission property

src/resources/swagger.yaml
jaserud: Yea this was a late day decision, didn't really think on it much so I just made a new router... I agree it should be in clinical will update :+1:

src/submission/submission-to-clinical/stat-calculator.ts
```diff
@@ -1,139 +1,197 @@
...
+type ForceRecaculateFlags = {
+  recalcEvenIfComplete?: boolean; // used to force recalculate if stat is already 100%
+  recalcEvenIfOverriden?: boolean; // used to force recalculate if previously overriden
 };
 
-// this function will reset stats and recalculate
-export const recalculateAllClincalInfoStats = (
-  donor: DeepReadonly<Donor>,
-  overrideSchema?: SchemasDictionary,
+const coreClinialEntities = new Set<CoreClinicalEntites>([
```
blabadi, line 15: I would define this in the clinical entities file

sampleFiles/clinical/specimen.tsv
jaserud: There were two of the same row, so I deleted one...

src/clinical/clinical-entities.ts
jaserud: Specimen can be partial since we want a record for each specimen. So it can be a decimal.

jaserud: I wanted to keep it consistent, also if any of them changes how core is calculated we may have to use numbers for it

sampleFiles/clinical/chemotherapy.tsv
jaserud: Hmm, guess it doesn't work with the sample schema or the ones in dev/qa eh. I was just working with this to create a file for int-test, must have committed it. Will revert since it doesn't match those schema

migrations/20200420215640-remove-clinical-info-stats.js
jaserud: Ah, was not aware. In that case I'll remove this migration script :+1:

src/submission/submission-api.ts
jaserud: :eyes: will fix when I move this to clinical-api

src/clinical/clinical-api.ts
```diff
@@ -37,6 +38,28 @@ class ClinicalController {
       .status(200)
       .send(id);
   }
+
+  @HasFullWriteAccess()
+  async patchDonorCompletenessStats(req: Request, res: Response) {
```
blabadi, line 43: just a question not related to code, if we update the same donor these manual values will be overridden, right ?

jaserud: yea, that's how it is setup

PR #474, opened 2020-04-27
Feat/467 lectern client array input

jaserud:
**Description of changes**
This PR adds support for array fields in schema
- Updated enum, range, & regex validation functions to handle array of values
- Added nonArrayInput validation for cases where array is submitted but shouldn't be
- added isArray designation chagne in change-analyzer
- Updated unit and integration tests to capture array inputs

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing


jaserud:
I'm putting in the PR request now, but we can merge after yours is merged.

sampleFiles/sample-schema.json
```diff
@@ -1158,6 +1158,7 @@
               "name": "presenting_symptoms",
               "description": "Indicate presenting symptoms at time of primary diagnosis.",
               "valueType": "string",
+              "isArray": true,
```
blabadi, line 1161: is this only for code lists ?

jaserud: Not necessarily, I accounted for regex and ranges too since I don't see why they also couldn't be arrays. Right now this is the only field that is going to be an array, but I added some other test cases in the unit tests for other fields types..

src/lectern-client/schema-entities.ts
```diff
@@ -3,14 +3,14 @@ import { DeepReadonly } from 'deep-freeze';
...
-  readonly [k: string]: string;
+  readonly [k: string]: string | string[];
 }
 
 export class TypedDataRecord {
   readonly [k: string]: SchemaTypes;
 }
 
-export type SchemaTypes = string | boolean | number | undefined;
+export type SchemaTypes = string | string[] | boolean | boolean[] | number | number[] | undefined;
```
blabadi, line 13: how can boolean be an array ?

jaserud: It could be if we set `isArray: true` in the schema, which is set at the field level. There is nothing in the schema that stops one from doing that. I agree it probably won't make sense in any of our contexts, but I just kept it general with the types since lectern client doesn't have any concepts of whether the input makes sense in the context.

blabadi:
no schema migrations tests where harmed in the process ?

src/lectern-client/change-analyzer.ts
```diff
@@ -90,13 +91,28 @@ export const analyzeChanges = (schemasDiff: SchemasDictionaryDiffs): ChangeAnaly
...
+          categorizeFieldArrayDesignationChange(analysis, field, fieldDiff.isArray);
+        }
       }
     }
   }
 
   return analysis;
 };
 
+const categorizeFieldArrayDesignationChange = (
```
blabadi, line 105: what if isArray is deleted ? or is it required in the field definition now ?

blabadi: I see you are checking only for not created or false, never mind

jaserud:
> no schema migrations tests where harmed in the process ?

So I tested migrating locally, and it works fine if a field is just one value since during submission it would've been the same in the converted record (we can't convert a value to array if it doesn't have the commas). The lectern-client will convert it to array before doing the validation where appropriate. 

The only thing here is that migration won't convert the field value of string to string[] (would need a manual script right now) which would happen in submission since we take the `processedRecord`. I did think about running the merge-function on the donor with the `processedRecords` returned from the schema processing, but decided that change is too big for this PR. Also I don't know how we feel about schema update process changing records in the donor,.

src/lectern-client/schema-entities.ts
blabadi: yeah I saw that in lectern.. so it's a feature ['true', 'Yes', 'Y', 'Si', 'Oui']

jaserud: Yea I guess. Lol could use it for psuedo binary inputs too eh [true, false, true, false] = [10101]

blabadi:
I haven't released the multiple PD stuff so up to you to wait for tag or merge now, I'll let you know when I finish dev test

PR #481, opened 2020-05-01
Fix sample schem and health

blabadi:
**Description of changes**

- [ ] Manual testing

### DIFF
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -114,6 +114,7 @@ const setJwtPublicKey = (keyUrl: string) => {
 };
 
 const setupRxNormConnection = (conf: RxNormDbConfig) => {
+  if (!conf.host) return;
   const pool = initPool({
     database: conf.database,
     host: conf.host,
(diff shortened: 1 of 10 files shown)


src/bootstrap.ts
```diff
@@ -176,6 +179,11 @@ export const run = async (config: AppConfig) => {
       throw new Error('App is not configured correctly either provide jwt pub key url or key');
     }
     setJwtPublicKey(config.jwtPubKeyUrl());
+    setInterval(() => {
```
jaserud, line 182: does the public key change every 5min now??

jaserud, line 182: does the public key change every 5min now??

src/bootstrap.ts
```diff
@@ -123,22 +124,24 @@ const setupRxNormConnection = (conf: RxNormDbConfig) => {
     timeout: conf.timeout,
   });
 
-  pool.on('error', err => setStatus('rxNormDb', { status: Status.ERROR }));
   pool.on('connection', () => setStatus('rxNormDb', { status: Status.OK }));
+
+  setStatus('rxNormDb', { status: Status.OK });
```
jaserud, line 129: won't this always be status `OK` now?? I don't see anywhere it would be set to error.

src/bootstrap.ts
blabadi: No this is just something should been done before to adapt automatically without manual restarts

jaserud: I see makes sense

src/bootstrap.ts
blabadi: yep, it was only checking on the beginning I tried set interval and ping the db but that didn't work.. that check missed up with the tests somehow. so for now I just want to get done with it, not urgent thing

jaserud: cool as long as we don't forget it.

blabadi:
> Looks like some tests fail eh. Running locally i got errors with rx db

yeah tests are still failing even without the set interval, I'm checking again, it's weird
I think something related to test container

blabadi:
turned out to be the wait on log message was no good, it tried to connect too early, should pass now

PR #482, opened 2020-05-04
Feature/download all clinical data

jaserud:
**Description of changes**

Relevant ticket: https://github.com/icgc-argo/platform-ui/issues/1318

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing


src/resources/swagger.yaml
```diff
@@ -602,6 +602,33 @@ paths:
           description: Active submission is not in PENDING_APPROVAL state and cannot be reopned
         '500':
           $ref: '#/components/responses/ServerError'
+  /submission/program/{programId}/clinical/committed/tsv:
```
blabadi, line 605: this is not a submission endpoint, it's in the clinical data domain
`/clinical/program/{programId}/tsv-export`

src/submission/submission-api.ts
```diff
@@ -259,6 +261,33 @@ class SubmissionController {
     const donorsJson = await fsPromises.readFile(req.file.path, 'utf-8');
     return res.status(201).send(await submission.operations.adminAddDonors(JSON.parse(donorsJson)));
   }
+
+  @HasProgramWriteAccess((req: Request) => req.params.programId)
+  async downloadCommittedClinicalDataAsTsv(req: Request, res: Response) {
```
blabadi, line 266: should be in clinical-api.ts

src/submission/submission-service.ts
```diff
@@ -641,6 +646,72 @@ export namespace operations {
...
+          clincialInfoRecords = generateSampleRegistrationRecordsFromDonor(d);
+        } else {
+          clincialInfoRecords = getClinicalEntitiesFromDonorBySchemaName(d, entity);
+        }
+        recordsMap[entity] = _.concat(recordsMap[entity] || [], clincialInfoRecords);
+      });
+    });
+
+    // map into object ready for api processing
+    const schemasWithFields = schemaManager.instance().getSchemasWithFields();
```
blabadi, line 670: do we actually need the schema? shouldn't we export whatever is in the clinicalInfo field ?

src/submission/submission-service.ts
```diff
@@ -641,6 +646,72 @@ export namespace operations {
...
+    const schemaVersion = schemaManager.instance().getCurrent().version;
+    return Object.entries(recordsMap)
+      .map(([schemaEntityName, records]) => {
+        if (isEmpty(records)) return undefined;
+
+        const relevantSchemaWithFields = schemasWithFields.find(s => s.name === schemaEntityName);
+        if (!relevantSchemaWithFields) {
+          throw new Error(`Can't find schema ${schemaEntityName}, shouldn't be possbile`);
```
blabadi, line 678: a schema can be deleted / renamed, this will fail then right ?

src/submission/submission-service.ts
```diff
@@ -641,6 +646,72 @@ export namespace operations {
...
+        }
+
+        return {
+          schemaEntityName: relevantSchemaWithFields.name,
+          schemaVersion,
```
blabadi, line 683: different donors can be on different versions

src/utils.ts
```diff
@@ -36,6 +36,33 @@ export namespace TsvUtils {
     });
     return rows.filter(notEmpty);
   };
+
+  export const parseObjectToTsv = (recordsObjects: any, headers: string[]) => {
```
blabadi, line 40: parse means reading a string, this is serialization / conversion

src/submission/submission-service.ts
```diff
@@ -641,6 +646,72 @@ export namespace operations {
     );
   };
 
+  export async function getAllCommittedClinicalData(programId: string) {
```
blabadi, line 649: can you profile this and see how much time it takes from the api for 4k & 8k of donors ?

src/submission/submission-service.ts
blabadi: that can cause some inconsistency in fields if the latest is different from the donor doc in some fields

src/submission/submission-service.ts
jaserud: So I did think about this a bit. But the donor version doesn't imply the clinical record is on the same version. Like they could have submitted and verified bu the donor schema version could be old

jaserud: Honestly I think this [ticket](https://github.com/icgc-argo/platform-ui/issues/1318) is misleading cause it feels more like a give me recrods that I can resubmit which is why I try to map them to the headers/fieldsNames in current schema

src/utils.ts
jaserud: correct, will rename this better :+1:

src/submission/submission-service.ts
jaserud: will do. Hmm but this function shouldn't be awaited I will switch it to be async.

blabadi: what do you mean should not be awaited ?

jaserud: right now I await on the getDonors query. Doesn't make sense to block and wait for mongo to return with the data. I think it makes more sense to do a `.then` callback instead

blabadi: not sure I understand, because `await` and `then` are same thing just different syntax to avoid call back hell

jaserud:
This PR will be outdated after dev is updated with refactor. I'll make a new one when ready.

PR #488, opened 2020-05-06
added common-model and refactoring to use it

jaserud:
**Description of changes**

**Type of Change**

- [ ] Bug
- [x] Refactor
- [ ] New Feature

- [ ] Manual testing


src/lectern-client/index.ts
```diff
@@ -1,5 +1,5 @@
-import * as schemaManager from '../submission/schema/schema-manager';
```
blabadi, line 1: I dont think this is used? lectern client should not have any dependency except for utils

src/resources/swagger.yaml
```diff
@@ -70,7 +70,7 @@ paths:
                   type: object
     patch:
       tags:
-        - Schemas
+        - Dictionary
```
blabadi, line 73: can we create a tag for dictionary migraiton only will be easier for admins ?

blabadi: can you also create a tag called Admin and tag all admin APIs like Patch Operations and migration, icgc import etc

blabadi: because I just noticed tags is an array so we can add many here

src/app.ts
```diff
@@ -66,9 +66,10 @@ app.use(
 app.use('/submission/configs', configRouter);
 app.use('/submission/program/:programId/registration', registrationRouter);
 app.use('/submission/program/:programId/clinical', submissionRouter);
-app.use('/submission/schema', schemaRouter);
```
blabadi, line 69: wont this be a breaking change for the apis used by gateway?  `list` `/template/*`

blabadi: I thought you wanted to deprecate the used ones first

src/resources/swagger.yaml
```diff
@@ -178,10 +178,10 @@ paths:
             application/json:
               schema:
                 type: object
-  /submission/schema/list:
+  /dictionary/list:
```
blabadi, line 181: I meant to open a ticket for gateway and some one will do that later, not for you to do it now you can keep the old path and just deprecate it

blabadi: even if we are going to do it right away it's still better to do it in two steps to avoid the overhead of coordinating deployments

src/lectern-client/index.ts
jaserud: correct, this was already there for some reason. will remove :+1:

src/app.ts
jaserud: hmm, yea I was thinking of making the changes right after merging but I guess thats not happening yet. I'll put it back in for now

src/resources/swagger.yaml
jaserud: Agree, it makes sense to keep then organized together :+1:

src/resources/swagger.yaml
jaserud: Agreed. I'm chaning in swagger though just because it will be messy to keep both and they are the same just with differnet path.

blabadi: swagger shows you deprecated if you mark them with deprecated

jaserud: yes but you have to keep the whole description. I can put it back if you prefer but it'll be the same things just different paths...

blabadi: but yeah no need to show them in swagger no one is checking them anyway

src/resources/swagger.yaml
```diff
@@ -785,15 +793,77 @@ paths:
...
+      summary: 'use `GET /dictionary/template/all` instead'
+      tags:
+        - Deprecated
+  /submission/schema/template/{schemaName}:
+    get:
+      deprecated: true
+      summary: 'use `GET /dictionary/template/{schemaName}` instead'
+      tags:
+        - Deprecated
+
```
jaserud, line 855: ^ Swagger-ui will render even if you don't specify the response body. I changed my mind, it is probly better to keep them for our own sanity so we can look up what is changed for what.

src/routes/dictionary.ts
```diff
@@ -1,25 +1,28 @@
...
 router.get('/template/:schemaName', wrapAsync(schemaApi.getTemplate));
 
+// dictionary migration
+router.post('/migration/dry-run-update', wrapAsync(schemaApi.schemaController.dryRunUpdate));
+router.post('/migration/run', wrapAsync(schemaApi.schemaController.update));
+router.post('/migration/resume', wrapAsync(schemaApi.schemaController.resumeMigration));
+router.get('/migration/:id', wrapAsync(schemaApi.schemaController.getMigration));
+router.get('/migration/', wrapAsync(schemaApi.schemaController.getMigration));
+
+// deprecated endpoints
```
jaserud, line 24: I decided to leave them. Not hurting any one and swagger has them defined so we'll know what to remove.

src/resources/swagger.yaml
jaserud: and also we'll know what to remove when its time to remove them.

blabadi: sounds good not a big deal

src/dictionary/api.ts
```diff
@@ -61,7 +61,7 @@ export const get = async (req: Request, res: Response) => {
   return res.status(200).send(schema);
 };
 
-export const getClinicalEntities = async (req: Request, res: Response) => {
+export const getClinicalEntitiesTemplate = async (req: Request, res: Response) => {
```
blabadi, line 64: this is the `/list` right ? it can be used for anything not just templates

jaserud: oh crap, your right I don't know why I renamed this... changing back

src/resources/swagger.yaml
```diff
@@ -55,10 +55,10 @@ paths:
         '200':
           description: 'Sample registration and Clinical Submission disabled state has been set'
 
-  /submission/schema:
+  /dictionary/migration/run:
     get:
```
blabadi, line 59: the get here is incorrect right ?

blabadi: before it was /dictionary/ and returns to you the current dictionary

src/routes/dictionary.ts
```diff
@@ -1,25 +1,28 @@
...
+router.patch('/', wrapAsync(schemaApi.schemaController.update));
```
blabadi, line 25: /dictionary/ should return the current dictionary

blabadi: it's not deprecated

src/resources/swagger.yaml
jaserud: your right it should be different endpoint will fix

blabadi:
looks good except for the comments, nothing else stands out

src/routes/dictionary.ts
jaserud: the patch is deprecated, the get is still their (at the top). the patch is being replaced by `migration/run`

jaserud:
Tests are passing
unit-test:
image

int-test:
image

PR #493, opened 2020-05-08
Feature/download all clinical data

jaserud:
**Description of changes**

Confinued from: https://github.com/icgc-argo/argo-clinical/pull/482

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature

- [x] Manual testing


blabadi:
we need a test for this it's a new feature

src/clinical/clinical-service.ts
```diff
@@ -128,3 +134,68 @@ export const updateDonorStats = async (donorId: number, coreCompletionOverride:
...
+      return {
+        entityName,
+        records,
+        entityFields: relevantSchemaWithFields?.fields || [],
+      };
+    })
+    .filter(notEmpty);
+
+  const end = new Date().getTime() / 1000;
+  L.info(`getClinicalData took ${end - start}s`);
```
blabadi, line 176: so how much time did it take for a bulk of donors ?

blabadi: this can also be debug statement not info

src/clinical/clinical-service.ts
```diff
@@ -128,3 +134,68 @@ export const updateDonorStats = async (donorId: number, coreCompletionOverride:
...
+    });
+  });
+
+  // map into object ready for api processing
+  const schemasWithFields = dictionaryManager.instance().getSchemasWithFields();
+  const data = Object.entries(recordsMap)
+    .map(([entityName, records]) => {
+      if (isEmpty(records)) return undefined;
+
+      const relevantSchemaWithFields = schemasWithFields.find(s => s.name === entityName);
```
blabadi, line 165: this should throw an error if the result is undefined

blabadi: so we can detect the issue

src/clinical/clinical-service.ts
jaserud: Running locally here's what I'm getting:
- 20,000 donors with sample registration take about 6 secs
- 3000 donors with sample registration is about 1 sec

I added donor clinical info in the 3000 case and it still took about 1 sec.

jaserud: Can't speak for dev/qa until I run their. Also the zip file sizes aren't too big compared, its in ~KB range.

blabadi: so it's blocking for 1 second at 3k

jaserud: yea, looks like it.

blabadi: my suggestion is to do this in a thread pool if time permits or open a ticket to convert it to thread pool because we don't know how much clinical data it will add, it's the right way to do it

jaserud:
looks like test is failing will take a look

src/clinical/donor-repo.ts
```diff
@@ -117,6 +121,25 @@ export const donorDao: DonorRepository = {
     return F(mapped);
   },
 
+  // Like findByProgramId, but DocQuery asks mongo to return PoJo without docIds for faster fetch
+  async findByProgramIdOmitMongoDocId(
```
jaserud, line 125: I found an optimization here. The download all takes about ~300ms for 3000 donors with sample registration and donor clinicalInfo.  I went a ahead and made this change, but we will probably need a more comprehensive look later.

blabadi: cool good to know this :+1:

src/clinical/service-worker-thread/runner.ts
```diff
@@ -0,0 +1,17 @@
...
+const poolSize = 1;
+const poolTimeOutMs = 20 * 1000;
+const jsWorkerLocation = __dirname + '/worker.js';
+
+const pool = new StaticPool({
+  size: poolSize,
+  task: jsWorkerLocation,
+});
+
+export async function runTaskInWorkerThread(taskToRun: WorkerTasks, taskArgs: any) {
```
blabadi, line 13: from what I understand this is designed to be generic right

jaserud: yes, that is the idea. I tried to keep it generic so when we add other data export functions they can be non blocking too.

src/clinical/donor-repo.ts
```diff
@@ -48,6 +48,10 @@ export interface DonorRepository {
     programId: string,
     projections?: Partial<Record<DONOR_FIELDS, number>>,
   ): Promise<DeepReadonly<Donor[]>>;
+  findByProgramIdOmitMongoDocId(
```
blabadi, line 51: can we instead use the same previous method with optional arguments to return lean object and use the new method you created as private method ?

src/clinical/donor-repo.ts
```diff
@@ -117,6 +121,25 @@ export const donorDao: DonorRepository = {
...
+  async findByProgramIdOmitMongoDocId(
+    programId: string,
+    projection?: Partial<Record<DONOR_FIELDS, number>>,
+  ): Promise<DeepReadonly<Donor[]>> {
+    const result = await DonorModel.find(
+      {
+        [DONOR_FIELDS.PROGRAM_ID]: programId,
+      },
+      projection,
+      { sort: { [DONOR_FIELDS.DONOR_ID]: 1 } },
```
blabadi, line 134: why are we sorting here ? just for my information

jaserud: Not sure. This was copy pasta from `findByProgramId` so it was already their. I think it's related to the DonorSchema.index

src/clinical/donor-repo.ts
jaserud: hmm, I don't see why not

src/clinical/service-worker-thread/runner.ts
```diff
@@ -0,0 +1,17 @@
+import { StaticPool } from 'node-worker-threads-pool';
+import { WorkerTasks } from './tasks';
+
+const poolSize = 1;
+const poolTimeOutMs = 20 * 1000;
```
blabadi, line 5: willl you wait 20 seconds for a download ?

src/clinical/service-worker-thread/runner.ts
```diff
@@ -0,0 +1,17 @@
+import { StaticPool } from 'node-worker-threads-pool';
+import { WorkerTasks } from './tasks';
+
+const poolSize = 1;
```
blabadi, line 4: why 1 ?

jaserud: Yea I was just thinking about this now. Maybe this should be configurable??

jaserud: No real reason for one. But I imagine under loads might need more then that

test/integration/clinical/stub-schema.json
```diff
@@ -118,6 +139,1390 @@
                   "polyA+ RNA",
                   "Other RNA fractions"
                 ]
+              },
```
blabadi, line 142: what happened here ?

jaserud: the schema was super outdated.... I just copied the stub ver 1.0 from submisison's stub sample-schema.json

src/clinical/service-worker-thread/runner.ts
jaserud: This if the worker thread function never returns anything, so the timeout will let us know to something went wrong. I guess this should be configurable too eh. I picked 20secs arbitrarily.

src/clinical/service-worker-thread/runner.ts
blabadi: it can be max the number of cores given, there is a config for lectern parallel, maybe you can repurpose that config

src/clinical/service-worker-thread/runner.ts
blabadi: I see I guess it's fine for now

src/clinical/service-worker-thread/runner.ts
jaserud: good point, we can just use that :+1:

test/integration/clinical/stub-schema.json
jaserud: as discussed I've moved the submission stub-schema to the /test/integration and used in both so shouldn't happen again.

blabadi:
squash reminder

PR #518, opened 2020-05-26
update cached dictionary if db version is different

jaserud:
**Description of changes**

Relevant changes are in the dictionary-manager
- updated getCurrent to update if versions don't match with db
- refactoring in dictionary-manager
- add a small test

**Type of Change**

- [x] Bug
- [x] Refactor
- [ ] New Feature

- [x] Manual testing


src/dictionary/manager.ts
```diff
@@ -20,27 +20,34 @@ const L = loggerFor(__filename);
...
-  private currentSchemaDictionary: SchemasDictionary = {
-    schemas: [],
-    name: '',
-    version: '',
-  };
+  constructor(private schemaServiceUrl: string, private dictionaryName: string) {}
 
-  constructor(private schemaServiceUrl: string) {}
+  getCurrent = async (projection?: SchemasDictionaryProjection): Promise<SchemasDictionary> => {
+    const dictionary = await schemaRepo.get(this.dictionaryName, projection);
```
blabadi, line 32: instead of always fetching the full schema whenever this method is called, as I mentioned in the ticket description you can only check the version to find if a newer version was loaded, and only then do a full load if needed. the schema document can be big, did that not work or what is your opinion?

src/dictionary/manager.ts
```diff
@@ -20,27 +20,34 @@ const L = loggerFor(__filename);
 
 let manager: SchemaManager;
 
+type SchemaWithFields = {
+  name: string;
+  fields: string[];
+};
+
 class SchemaManager {
-  private currentSchemaDictionary: SchemasDictionary = {
```
blabadi, line 24: as I mentioned below, you can keep this as cached version instead of doing a full load for the dictionary document every time. only send the cached version number and check if it was changed, if yes, do a load if not, return this cached version

src/dictionary/manager.ts
jaserud: hmm, agreed. I didn't see that in the issue and I don't see any issues with doing it that way. Will update :+1:

src/dictionary/manager.ts
```diff
@@ -29,18 +34,38 @@ class SchemaManager {
...
   constructor(private schemaServiceUrl: string) {}
 
-  getCurrent = (): SchemasDictionary => {
+  getCurrent = async (): Promise<SchemasDictionary> => {
+    await this.updateManagerDitionaryIfNeeded();
     return this.currentSchemaDictionary;
   };
 
-  getSchemasWithFields = (
+  updateManagerDitionaryIfNeeded = async () => {
```
blabadi, line 42: typo in Dictionary

jaserud:
> I have couple of tickets in dev testing I'll need to test them in dev and RC

yea no rush. I'll merge this after rc.

PR #577, opened 2020-07-22
abort migration when diction cannot be found in lectern

Buwujiu:
**Description of changes**

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [x] Manual testing

### DIFF
--- a/src/dictionary/manager.ts
+++ b/src/dictionary/manager.ts
@@ -190,7 +190,7 @@ class SchemaManager {
   loadSchemaByVersion = async (
     name: string,
     version: string,
-  ): Promise<dictionaryEntities.SchemasDictionary | undefined> => {
+  ): Promise<dictionaryEntities.SchemasDictionary> => {
     try {
       const newSchema = await dictionaryRestClient.fetchSchema(
         this.schemaServiceUrl,
@@ -200,7 +200,7 @@ class SchemaManager {
       return newSchema;
     } catch (err) {
       L.error('Failed to fetch schema: ', err);
-      return undefined;
+      throw new Error('Failed to fetch schema: ' + err.message);
     }
   };
 
--- a/src/dictionary/manager.ts
+++ b/src/dictionary/manager.ts
@@ -199,8 +199,8 @@ class SchemaManager {
       );
       return newSchema;
     } catch (err) {
-      L.error('Failed to load fetch schema', err);
-      return Promise.resolve(undefined);
+      L.error('Failed to fetch schema: ', err);
+      return undefined;
     }
   };
 
--- a/src/dictionary/manager.ts
+++ b/src/dictionary/manager.ts
@@ -176,6 +176,9 @@ class SchemaManager {
     newVersion: string,
   ): Promise<dictionaryEntities.SchemasDictionary> => {
     const newSchema = await this.loadSchemaByVersion(name, newVersion);
+    if (newSchema == undefined) {
+      throw new Error("couldn't save/update new schema, schema is undefined.");
+    }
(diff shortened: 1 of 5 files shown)


src/dictionary/manager.ts
```diff
@@ -187,9 +190,10 @@ class SchemaManager {
   loadSchemaByVersion = async (
     name: string,
     version: string,
-  ): Promise<dictionaryEntities.SchemasDictionary> => {
-    const newSchema = await dictionaryRestClient.fetchSchema(this.schemaServiceUrl, name, version);
-    return newSchema;
+  ): Promise<dictionaryEntities.SchemasDictionary | undefined> => {
+    return await dictionaryRestClient
+      .fetchSchema(this.schemaServiceUrl, name, version)
+      .catch(err => undefined);
```
blabadi, line 196: when you are using await catch is not the best to use, you should use `try / catch`, log the error and decide to return or re throw the error

src/dictionary/manager.ts
```diff
@@ -187,9 +190,18 @@ class SchemaManager {
...
+  ): Promise<dictionaryEntities.SchemasDictionary | undefined> => {
+    try {
+      const newSchema = await dictionaryRestClient.fetchSchema(
+        this.schemaServiceUrl,
+        name,
+        version,
+      );
+      return newSchema;
+    } catch (err) {
+      L.debug(err.message);
```
blabadi, line 202: L.error not debug

src/dictionary/manager.ts
```diff
@@ -187,9 +190,18 @@ class SchemaManager {
...
+      return Promise.resolve(undefined);
```
blabadi, line 203: you can return undefined without Promise.resolve (no need to use promises with async/await) 

that said, I think it's better to rethrow the error in this case because we don't know why it's failed

Buwujiu: If we throw error here, migration won't be aborted later which makes the current invalid migration in `open` status, instead we want migration to be `closed`, so still need to return `undefined` here.

blabadi: you can try catch there and abort the migration

Buwujiu: this is what we want: [
Screen Shot 2020-07-23 at 3 51 49 PM
](url)

src/submission/migration/migration-manager.ts
```diff
@@ -139,6 +139,10 @@ export namespace MigrationManager {
       newSchemaVersion,
```
blabadi, line 139: in this line surround this with try catch in the catch you abort the migration. also it would be nice if the migration stored the reason it failed in such case

src/submission/migration/migration-manager.ts
```diff
@@ -134,10 +134,15 @@ export namespace MigrationManager {
...
-      newSchemaVersion,
-    );
+    let newTargetSchema: dictionaryEntities.SchemasDictionary;
+    try {
+      newTargetSchema = await dictionaryManagerInstance().loadSchemaByVersion(
+        dictionaryName,
+        newSchemaVersion,
+      );
+    } catch (err) {
+      return await abortMigration(migrationToRun);
```
blabadi, line 144: so `abortMigration` takes a 2nd argument where it stores the failure analysis right, could you modify it so it also accepts a message like `schema loading failed, check that version is correct or try again`

blabadi: or any other way to indicate the reason of abortion

Buwujiu: Screen Shot 2020-07-24 at 9 43 41 AM

PR #604, opened 2020-11-04
Dependency updates

hlminh2000:


blabadi (approved):

hlminh2000:
btw @blabadi, FYI the pre-push hook doesn't work on my mac because `grep` on mac isn't the gnu version...
So on my machine, I had to do `brew install grep`, then change the hook to the following for it to work:
```
git branch | ggrep \"*\" | ggrep -E -v \"^* develop$\" && git branch | ggrep \"*\" | ggrep -E -v \"^* master$\"
```
Not sure if anyone else working on this repo has this problem, so I didn't look into a long-term solution, but reporting so you're aware...

It looks like the script is trying to protect `develop` and `master, which is probably better done through github protected branch...

blabadi:
> btw @blabadi, FYI the pre-push hook doesn't work on my mac because `grep` on mac isn't the gnu version...
> So on my machine, I had to do `brew install grep`, then change the hook to the following for it to work:
> 
> ```
> git branch | ggrep \"*\" | ggrep -E -v \"^* develop$\" && git branch | ggrep \"*\" | ggrep -E -v \"^* master$\"
> ```
> 
> Not sure if anyone else working on this repo has this problem, so I didn't look into a long-term solution, but reporting so you're aware...
> 
> It looks like the script is trying to protect `develop` and `master, which is probably better done through github protected branch...

created: https://github.com/icgc-argo/argo-clinical/issues/607

PR #663, opened 2021-04-21
feature to allow schema field value type change from integer to number

Buwujiu:
**Description of changes**

This PR implements a new feature to allow schema field's `valueType` to change from integer to number in dictionary migration. 

There are two steps to achieve this: 
1. Use lectern client's `fetchDiff` endpoint to get dictionary migration diffs.
2. For each clinical entity, go through the `valueTypeChanges` array and exam if there is integer to number value type change, if there is, allow and only allow this change, other types of value type changes are prohibited. 

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/submission/migration/migration-manager.ts
```diff
@@ -310,18 +314,31 @@ export namespace MigrationManager {
...
         const field: string = change.split('.')[1];
-        changedFields.push(field);
+        prohibitedChangedFields.push(field);
       }
     });
-    return changedFields;
+    return prohibitedChangedFields;
+  }
+
+  function allowValueTypeUpdate(diffs: SchemasDictionaryDiffs, valueTypeChange: string): boolean {
```
Buwujiu, line 332: This function is the main point of the feature, it checks the `diff` of this value type change and verifies `before` and `after` value type to allow change.

src/submission/migration/migration-manager.ts
```diff
@@ -257,6 +260,7 @@ export namespace MigrationManager {
       );
 
       const changedValueTypes: string[] = checkValueTypeChanges(
+        diff,
```
Buwujiu, line 263: Have to call `fetchDiff` to get differences of 2 dictionaries because `valueTypeChanges` only contain names of changed fields without info of what's being changed, therefor `valueTypeChanges` alone does not provide enough information to determine if the value type is changed from integer to number.

src/submission/migration/migration-manager.ts
```diff
@@ -236,6 +237,8 @@ export namespace MigrationManager {
 
     const analysis = await dictionaryManagerInstance().analyzeChanges(currVersion, toVersion);
 
+    const diff = await dictionaryManagerInstance().fetchDiff(currVersion, toVersion);
```
blabadi, line 240: I think what we should do is update the analyze changes method instead of fetching diff again
it's already fetched in the analysis step

src/submission/migration/migration-manager.ts
blabadi: then you can change valueTypeChanges to include what you need, did you try that?

src/submission/migration/migration-manager.ts
```diff
@@ -312,16 +316,32 @@ export namespace MigrationManager {
   function checkValueTypeChanges(
     valueTypeChanges: string[],
     clinicalEntityName: ClinicalEntitySchemaNames,
+    currentSchema: dictionaryEntities.SchemasDictionary,
+    newSchemaEntity: dictionaryEntities.SchemaDefinition | undefined,
```
blabadi, line 320: this should not allow `undefined`

src/submission/migration/migration-manager.ts
```diff
@@ -312,16 +316,32 @@ export namespace MigrationManager {
...
   ): string[] {
-    const changedFields: string[] = [];
+    const prohibitedChangedFields: string[] = [];
     valueTypeChanges.forEach(change => {
       const changedSchema = change.split('.')[0];
       if (clinicalEntityName === changedSchema) {
         const field: string = change.split('.')[1];
-        changedFields.push(field);
+        const changedSchema_before = currentSchema.schemas.find(s => s.name === changedSchema);
```
blabadi, line 327: can you use camelCase instead of `_`

src/dictionary/manager.ts
```diff
@@ -171,6 +171,16 @@ class SchemaManager {
     return result;
   };
 
+  fetchDiff = async (oldVersion: string, newVersion: string) => {
```
blabadi, line 174: Is this still needed ?

Buwujiu: it's not used right now, but I thought maybe it can be used in the future so I kept it

src/submission/migration/migration-manager.ts
```diff
@@ -312,16 +316,34 @@ export namespace MigrationManager {
...
+        const changedSchemaBefore = currentSchema.schemas.find(s => s.name === changedSchema);
+        const changedFieldBefore = changedSchemaBefore?.fields.find(f => f.name === field);
+        const changedValueTypeBefore = changedFieldBefore?.valueType;
+
+        const changedFieldAfter = newSchemaEntity?.fields.find(f => f.name === field);
+        const changedValueTypeAfter = changedFieldAfter?.valueType;
+        // Only allow valut type change fron intger to number
+        if (
+          !(
+            changedValueTypeBefore &&
```
blabadi, line 336: if these are undefined we should throw error

src/submission/migration/migration-manager.ts
```diff
@@ -312,16 +316,34 @@ export namespace MigrationManager {
...
+            changedValueTypeAfter &&
```
blabadi, line 337: if these are undefined we should throw error

src/submission/migration/migration-manager.ts
```diff
@@ -312,16 +316,37 @@ export namespace MigrationManager {
...
-        changedFields.push(field);
+        const changedSchemaBefore = currentSchema.schemas.find(s => s.name === changedSchema);
+        const changedFieldBefore = changedSchemaBefore?.fields.find(f => f.name === field);
+        const changedValueTypeBefore = changedFieldBefore?.valueType;
+
+        const changedFieldAfter = newSchemaEntity?.fields.find(f => f.name === field);
+        const changedValueTypeAfter = changedFieldAfter?.valueType;
+        if (!(changedValueTypeBefore && changedValueTypeAfter)) {
+          const msg = `Field ${field} in schema ${changedSchema} has value type change, but field or schema are missing in current and new dictionary.`;
+          L.error(msg, new Error(msg));
```
blabadi, line 335: throw an error (throw new Error())

src/submission/migration/migration-manager.ts
blabadi: remove `| undefined`

blabadi (approved):
looks good to me, tests are still failing, you can merge when all is green :green_circle:

PR #765, opened 2022-04-27
comment out surgery until dictionary is ready

blabadi:


ciaranschutte (approved):
looks like all my chances reverted - thank you

PR #784, opened 2022-06-09
📊 #2259 Core Completeness updates

demariadaniel:
**Description of changes**

Clinical Service changes related to Submitted Data Core Completeness UI
* Updates Core Completeness to include Normal Specimen / Tumour Specimen Data
* Moves Core Completeness into Donor entity data
* Fixes problems in dev/qa: Missing Treatment entity data, Filters empty entries, Fixes specimen/sample registration data

Related API PR https://github.com/icgc-argo/platform-api/pull/565
Related UI PR https://github.com/icgc-argo/platform-ui/pull/2329

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [x] Manual testing


src/clinical/service-worker-thread/tasks.ts
```diff
@@ -71,27 +78,67 @@ const sortDocs = (sort: string) => (currentRecord: ClinicalInfo, nextRecord: Cli
...
+const mapEntityDocuments = (
+  totalDonors: number,
+  schemas: any,
+  query: ClinicalQuery,
+  completionStats: CompletionRecord[],
+) => (
+  entity: [string, ClinicalInfo[]],
+  index: number,
+  originalResultsArray: [string, ClinicalInfo[]][],
+) => {
```
joneubank, line 90: Since this method is a bit complicated it would benefit from some comments explaining the work that is done in it and what the objective is. The return type of the inner method would be nice to declare as well.

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -71,27 +78,67 @@ const sortDocs = (sort: string) => (currentRecord: ClinicalInfo, nextRecord: Cli
...
-  results,
-]: [string, ClinicalInfo[]]) => {
-  if (isEmpty(results)) return undefined;
+const mapEntityDocuments = (
+  totalDonors: number,
+  schemas: any,
+  query: ClinicalQuery,
+  completionStats: CompletionRecord[],
+) => (
+  entity: [string, ClinicalInfo[]],
```
joneubank, line 87: Why do entity as a tuple instead of a declared type like:
```
type EntityClinicalInfo = {
  entityName: string;
  clinicalInfo: ClinicalInfo[];
}
```

No performance difference but it will be a lot easier to keep track of the meaning of the two variables when they're named.

src/common-model/functions.ts
```diff
@@ -133,27 +133,60 @@ export function getClinicalEntitySubmittedData(
...
+  // Sample Registration requires Specimen data
+  (entityTypes.includes('sampleRegistration') && !entityTypes.includes('specimens'))
+    ? ['completionStats', 'sampleRegistration', 'specimens']
+    : // Clinical Therapies require Treatments
+    // hormoneTherapy + treatment do not match schema names
+    entityTypes.includes('hormoneTherapy') ||
+      entityTypes.includes('treatment') ||
+      ClinicalTherapySchemaNames.some(entity => entityTypes.includes(entity))
+    ? ['treatments', 'treatment']
+    : '';
```
joneubank, line 188: Feels like the nested ternaries are getting a bit hard to parse... you might consider moving all the comments into a single block, and this might actually be a good use for an if/else structure since it would separate the conditions.

joneubank (approved):
Looks good, comments are just thoughts on some code style/readability that are my personal preferences.

src/clinical/service-worker-thread/tasks.ts
joneubank: Oh this is from the `Object.entries`, that explains it. I think the typing you've done here is fine, this makes more sense in context. If you add some comments on the method I think thats sufficient.

src/clinical/service-worker-thread/tasks.ts
demariadaniel: I ended up refactoring this a fair bit with comments and it should be much more readable

src/clinical/service-worker-thread/tasks.ts
demariadaniel: I ended up using an Object type + removing Object.entries which I'd been thinking of doing for a while. Also cleaned up the map function with new types added.

src/common-model/entities.ts
```diff
@@ -39,7 +39,7 @@ export enum ClinicalEntitySchemaNames {
   BIOMARKER = 'biomarker',
 }
 
-export const aliasEntityNames = {
+export const aliasEntityNames: Record<string, string> = {
```
blabadi, line 42: it's better to utilize typescript type system than using strings
for example instead of `Record<string, string>` use `Record<EntityName, EntityAlias>` where each one of these is either an enum or literal type `type EntityAlias = 'donor' | 'sampleRegistration' | ....` or  use `keyof ...` 
the point is to avoid Strings in places where arbitrary strings can cause runtime errors

src/common-model/functions.ts
```diff
@@ -133,27 +133,65 @@ export function getClinicalEntitySubmittedData(
...
             ...entity.clinicalInfo,
-          },
-    )
-    .filter(notEmpty);
+          }));
 
   return clinicalRecords;
 }
 
+export const requiredEntities = (entityTypes: string[]) => {
```
blabadi, line 176: same as above comment, also applies to return type

src/common-model/entities.ts
demariadaniel: Nice I've made these updates, definitely better

src/common-model/functions.ts
```diff
@@ -133,27 +133,65 @@ export function getClinicalEntitySubmittedData(
...
+    return ['completionStats', 'sampleRegistration', 'specimens'];
+  } else if (
+    // Clinical Therapies require Treatments
+    // hormoneTherapy + treatment do not match schema names
+    entityTypes.includes('hormoneTherapy') ||
+    entityTypes.includes('treatment') ||
+    ClinicalTherapySchemaNames.some(entity => entityTypes.includes(entity))
+  ) {
+    return ['treatments', 'treatment'];
+  } else return '';
```
blabadi, line 193: it's better to keep the return type consistent instead of empty string why not return an empty array

src/clinical/donor-repo.ts
```diff
@@ -190,7 +183,11 @@ export const donorDao: DonorRepository = {
   ): Promise<DeepReadonly<{ donors: Donor[]; totalDonors: number }>> {
     const { sort, pageSize, entityTypes, donorIds, submitterDonorIds, completionState } = query;
 
-    const projection = [...DONOR_ENTITY_CORE_FIELDS, ...entityTypes].join(' ');
+    const projection = [
+      ...DONOR_ENTITY_CORE_FIELDS,
+      ...entityTypes,
+      ...requiredEntities(entityTypes),
```
blabadi, line 189: ok so I see this function is to get the required document fields for the projection, I'd call it `getRequiredDonorDocumentFieldsForEntityTypes()` a bit more verbose but clearer

src/common-model/functions.ts
```diff
@@ -133,27 +134,67 @@ export function getClinicalEntitySubmittedData(
...
+    return ['completionStats', 'sampleRegistration', 'specimens'];
+  } else if (
+    // Clinical Therapies require Treatments
+    // hormoneTherapy + treatment do not match schema names
+    entityTypes.includes('hormoneTherapy') ||
+    entityTypes.includes('treatment') ||
+    ClinicalTherapySchemaNames.some(entity => entityTypes.includes(entity))
+  ) {
+    return ['treatments'];
+  } else return [''];
```
blabadi, line 195: why not just return `[ ]` the empty string is not really needed, right ?

demariadaniel: Yes updating that next commit

src/common-model/functions.ts
```diff
@@ -133,27 +134,67 @@ export function getClinicalEntitySubmittedData(
...
-    )
-    .filter(notEmpty);
+          }));
 
   return clinicalRecords;
 }
 
+export const requiredEntities = (
+  entityTypes: EntityAlias[],
+): EntityAlias | ClinicalEntitySchemaNames | string[] => {
```
blabadi, line 179: the return type can be simpler:
`Array<EntityAlias | 'completionStats'>`

src/clinical/donor-repo.ts
demariadaniel: Sure I'll use `getRequiredDonorFieldsForEntityTypes` just to keep it a little concise, I think 'document' is understood

src/common-model/functions.ts
demariadaniel: Updated

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -19,22 +19,47 @@
...
 type RecordsMap = {
   [key in ClinicalEntitySchemaNames]: ClinicalInfo[];
 };
 
+type EntityClinicalInfo = {
+  entityName: ClinicalEntitySchemaNames;
+  results: ClinicalInfo[];
+};
+
+const queryEntityNames = <string[]>Object.values(aliasEntityNames);
```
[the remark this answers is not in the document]
demariadaniel: Nice, updated

src/common-model/functions.ts
demariadaniel: Updated to `Array<EntityAlias | ClinicalEntitySchemaNames | 'completionStats'>`

Issue #820, opened 2022-10-04
Clinical API should accept API Keys from Ego


sifavahora:
Clinical API auth checks currently only accept Ego JWTs, they should also accept API Keys from Ego.
If the Authorization header on a request contains a UUID instead of a JWT, this should be considered an API Key. To check if this provides the user permission to access the requested data clinical needs to:

- send this API Key to Ego Check API Key endpoint, clinical will receive back a list of user permissions (same as in scopes of JWT): https://ego.icgc-argo.org/api/swagger-ui.html#/Api%20Keys/checkApiKeyUsingPOST
- clincial can use these permissions to check if user can be granted access (same rules as for  permissions from JWT)

lindaxiang:
Test all API Get endpoints for [Clinical Data](https://clinical.qa.argo.cancercollaboratory.org/api-docs/#/Clinical%20Data), all can use API keys as token.

edsu7:
List of GET points that failed on staging
```
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Submission%20Configurations/get_submission_configs
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Dictionary/get_dictionary_changes
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Sample%20Registration/get_submission_program__programId__registration:~:text=GET-,/submission/program/%7BprogramId%7D/registration,-Get%20active%20registration
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Submission/get_submission_program__programId__clinical_
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/get_clinical_program__programId__donor__donorId_
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/get_clinical_program__programId_clinical_data
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/get_clinical_program__programId_clinical_errorsget_clinical_program__programId__clinical_tsv
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/get_clinical_program__programId__tsv_all
https://clinical.platform.staging.icgc-argo.org/api-docs/#/Clinical%20Data/get_clinical_program__programId__tsv__entityType_
```

lindaxiang:
Save observation as what Edmund listed above.

[...]

PR #823, opened 2022-10-12
🔑  #820 Accept Ego API Keys

demariadaniel:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [x] Manual testing
- [x] Set `validationDependency` in meta tag for [Argo Dictionary](https://github.com/icgc-argo/argo-dictionary) fields used in code

### DIFF
--- a/src/decorators/index.ts
+++ b/src/decorators/index.ts
@@ -19,9 +19,12 @@
 
 import { Request, Response, RequestHandler, NextFunction } from 'express';
 import jwt from 'jsonwebtoken';
+import fetch from 'node-fetch';
 import { loggerFor } from '../logger';
 import { config } from '../config';
 const L = loggerFor(__filename);
+
+const EGO_URL = 'https://ego.dev.argo.cancercollaboratory.org/api/';
 
 const getToken = (request: Request) => {
   if (!request.headers.authorization) {
@@ -30,8 +33,11 @@ const getToken = (request: Request) => {
 
   const token = request.headers.authorization.includes('Bearer')
     ? decodeAndVerify(request.headers.authorization.split(' ')[1])
-    : verifyEgoApiKey(request.headers.authorization);
-
+    : verifyEgoApiKey('e4a841ef-cd62-412f-998c-0a3d985c4baf', request.headers.authorization);
+  console.log(
+    '\nverify',
+    verifyEgoApiKey('e4a841ef-cd62-412f-998c-0a3d985c4baf', request.headers.authorization),
+  );
   return token;
 };
 
@@ -49,7 +55,14 @@ const decodeAndVerify = (tokenJwtString: string) => {
   }
 };
 
-const verifyEgoApiKey = (uuidString: string) => {};
+const verifyEgoApiKey = async (uuidString: string, auth: string) => {
+  const response = await fetch(`${EGO_URL}/o/check_api_key?apiKey=${uuidString}`, {
+    method: 'post',
+    headers: { 'Content-Type': 'application/json', authorization: auth },
+  });
+  const jsr = await response.json();
+  console.log('response', jsr);
+};
 
 const hasScope = (scopes: string[], token: any) => {
   if (
--- a/src/decorators/index.ts
+++ b/src/decorators/index.ts
@@ -24,9 +24,6 @@ import { loggerFor } from '../logger';
 import { config } from '../config';
 const L = loggerFor(__filename);
 
-// TODO: MAKE ENV VARIABLE
-const EGO_URL = 'https://ego.dev.argo.cancercollaboratory.org/api/';
-
 const getToken = async (request: Request) => {
   if (!request.headers.authorization) {
     return undefined;
@@ -54,6 +51,7 @@ const decodeAndVerify = (tokenJwtString: string) => {
 };
 
 const verifyEgoApiKey = async (uuidString: string, auth: string) => {
+  const EGO_URL = config.getConfig().egoUrl();
   const response = await fetch(`${EGO_URL}/o/check_api_key?apiKey=${uuidString}`, {
     method: 'post',
     headers: { 'Content-Type': 'application/json', authorization: auth },
--- a/src/decorators/index.ts
+++ b/src/decorators/index.ts
@@ -73,7 +73,7 @@ const verifyEgoApiKey = async (keyString: string): Promise<TokenValidationResult
     .then(async res => {
       const token = await res.json();
       const { scopes } = token;
-      if (token.error) throw token.error;
+      if (token.error || !Boolean(scopes)) throw token.error || 'No token scopes provided';
       return { success: true, data: { scopes } };
     })
     .catch(err => {
--- a/src/decorators/index.ts
+++ b/src/decorators/index.ts
@@ -29,7 +29,7 @@ type TokenValidationResult = { success: boolean; data?: TokenData };
 type TokenData = {
   scopes: string[];
 };
-type JwtPayload = { context: { scopes: string[] } };
+type JwtPayload = { context: { scope: string[] } };
 
 const getToken = async (request: Request): Promise<TokenValidationResult> => {
   if (!request.headers.authorization?.startsWith('Bearer ')) {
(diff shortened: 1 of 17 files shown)


src/submission/submission-repo.ts
```diff
@@ -24,7 +24,7 @@ import { ActiveClinicalSubmission, SUBMISSION_STATE } from './submission-entitie
 import { MongooseUtils, F, Errors } from '../utils';
 import { InternalError } from './errors';
 import _ from 'lodash';
-import uuid from 'uuid';
+import { v1 as uuidv1 } from 'uuid';
```
demariadaniel, line 27: Use V4

src/decorators/index.ts
```diff
@@ -24,22 +24,21 @@ import { loggerFor } from '../logger';
...
   }
 
-  const token = request.headers.authorization.includes('Bearer')
-    ? decodeAndVerify(request.headers.authorization.split(' ')[1])
-    : await verifyEgoApiKey(request.headers.authorization, request.headers.authorization);
+  const authToken = request.headers.authorization.split(' ')[1];
+
+  const token = Boolean(jwt.decode(authToken))
+    ? verifyJwt(authToken)
+    : await verifyEgoApiKey(authToken);
```
joneubank, line 36: We still need a check somewhere that this header starts with "Bearer"
```suggestion
  if (!request.headers.authorization.startsWith('Bearer ') {
    return undefined;
  }

  const authToken = request.headers.authorization.slice(7); // remove 'Bearer ' from header

  const token = Boolean(jwt.decode(authToken))
    ? verifyJwt(authToken)
    : await verifyEgoApiKey(authToken);
```

src/decorators/index.ts
```diff
@@ -45,25 +52,42 @@ const decodeAndVerify = (tokenJwtString: string) => {
...
-    token.context.scope.filter((s: string) => scopes.indexOf(s) >= 0).length === 0
+    (token.context &&
+      (!token.context.scope ||
+        token.context.scope.filter((s: string) => scopes.indexOf(s) >= 0).length === 0)) ||
+    (token.scope && token.scope.filter((s: string) => scopes.indexOf(s) >= 0).length === 0)
   ) {
     return false;
   }
   return true;
 };
```
joneubank, line 79: I'm suggesting a re-write of this method so that it no longer accepts `any` as an input. Making this accept just the provided scopes from the token makes its logic very straightforward, and pushes all teh token parsing into the code that is already handling tokens.

```TypeScript

/**
 * Confirms if the at least one of the required scopes has been provided
 * @param requiredScopes at least one of these scopes must be provided
 * @param providedScopes scopes provided in auth header
 * @returns {boolean} true if provided scopes contains at least one of the requiredScopes
 */
const hasScope = (requiredScopes: string[], providedScopes: string[]): boolean => 
  requiredScopes.some(scope => providedScopes.includes(scope));
  
  ```

File: src/decorators/index.ts [hunk 4]
```diff
@@ -19,19 +19,26 @@
 
 import { Request, Response, RequestHandler, NextFunction } from 'express';
 import jwt from 'jsonwebtoken';
+import fetch from 'node-fetch';
 import { loggerFor } from '../logger';
 import { config } from '../config';
 const L = loggerFor(__filename);
 
-const getToken = (request: Request) => {
+const getToken = async (request: Request) => {
```
joneubank, line 27: I suggest changes below to `hasScopes` that require us to provide just the scopes form the token. To handle that, I would change the return type of this method from `Promise<any>` to `Promise<TokenValidationResult>` which uses some new types:

```TypeScript
type Result<T> = {success: false, message: string} | {success: true, data: T};
type TokenData = {
  scopes: string[];
  // ... more token properties if required
}
type TokenValidationResult = Result<TokenData>;
```

Note that `Result<T>` is a useful type that should probably be added in the utils file.

joneubank: You'll also need to modify the two verify methods (`verifyJwt`, `verifyEgoApiKey`) to return just the scopes array.

src/submission/submission-repo.ts
```diff
@@ -24,7 +24,7 @@ import { ActiveClinicalSubmission, SUBMISSION_STATE } from './submission-entitie
 import { MongooseUtils, F, Errors } from '../utils';
 import { InternalError } from './errors';
 import _ from 'lodash';
-import uuid from 'uuid';
+import { v4 as uuidv4 } from 'uuid';
```
joneubank, line 27: 🙏

joneubank (changes requested):
Looks functionally complete. I suggested some changes to reorganize some of the code.

src/decorators/index.ts
joneubank: I would actually add a trim to line 36 here to deal with extra whitespace issues:
` const authToken = request.headers.authorization.slice(7).trim();`

demariadaniel: Updated

src/decorators/index.ts
demariadaniel: Updated

src/decorators/index.ts
```diff
@@ -44,36 +48,37 @@ const verifyJwt = (tokenJwtString: string): TokenValidationResult => {
...
     method: 'post',
     headers: { 'Content-Type': 'application/json', authorization: auth },
   })
     .then(async res => {
       const token = await res.json();
-
+      const { scopes } = token;
       if (token.error) throw token.error;
-      return { success: true, data: token };
+      return { success: true, data: { scopes } };
```
joneubank, line 77: what if scopes is undefined?

joneubank (changes requested):

src/decorators/index.ts
demariadaniel: I'll update conditional `if (token.error || !Boolean(scopes))`

src/decorators/index.ts
```diff
@@ -18,52 +18,90 @@
...
 const L = loggerFor(__filename);
 
-const getToken = (request: Request) => {
-  if (!request.headers.authorization) {
-    return undefined;
+type TokenValidationResult = { success: boolean; data?: TokenData; error?: any };
+type TokenData = {
+  scopes: string[];
+};
+type JwtPayload = { context: { scopes: [] } };
```
joneubank, line 32: scopes should be `string[]`

src/decorators/index.ts
```diff
@@ -18,52 +18,90 @@
...
 import jwt from 'jsonwebtoken';
+import fetch from 'node-fetch';
 import { loggerFor } from '../logger';
 import { config } from '../config';
 const L = loggerFor(__filename);
 
-const getToken = (request: Request) => {
-  if (!request.headers.authorization) {
-    return undefined;
+type TokenValidationResult = { success: boolean; data?: TokenData; error?: any };
```
joneubank, line 28: i would probably restrict the types allowed as "error", no reason to let errors by any type

demariadaniel: I ended up removing error b/c it isn't used up the chain or returned in the request
So I just added logging wherever errors are captured and removed it from the type

src/decorators/index.ts
demariadaniel: Updated

src/decorators/index.ts
demariadaniel: Updated

src/decorators/index.ts
demariadaniel: Updated

src/decorators/index.ts
joneubank: `!Boolean(scopes)` is fine, I usually see this written like `!!scopes`

joneubank: `!scopes` might be sufficient. why do you use the `Boolean()` function?

demariadaniel: Eh just been using Boolean + other primitives a little more frequently after some recent code review suggestions, in this case it really is just stylistic. Updated conditional to `!scopes`

PR #828, opened 2022-10-26
Feat/#810 endpoints

ciaranschutte:
**Description of changes**

adds program exception service to api

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [x] Manual testing


src/exception/exception-repo.ts
```diff
@@ -17,102 +17,73 @@
...
-const ProgramExceptionSchema = new mongoose.Schema({
-  programName: { type: String, unique: true, required: true },
-  exceptions: { type: [ExceptionSchema], required: true },
+const programExceptionSchema = new mongoose.Schema({
+  programId: String,
+  exceptions: [
+    {
+      schema: String,
+      coreField: String,
+      exceptionValue: { type: String, enum: ['Unknown', 'Missing', 'Not applicable'] },
```
demariadaniel, line 33: Worth capturing this enum as a variable, likely going to have a future use

src/exception/exception-service.ts
```diff
@@ -0,0 +1,148 @@
...
+    return [
+      createValidationError(idx, `core field of ${record.requested_core_field} is not valid`),
+    ];
+  }
+
+  return [];
+};
+
+const checkRequestedValue = (record: ProgramExceptionRecord, idx: number) => {
+  const validRequests = ['Missing', 'Unknown', 'Not Applicable'];
```
demariadaniel, line 132: i.e. ExceptionValue enum would be used here

ciaranschutte:
cleaned up endpoints and added any missing types for "any"s @joneubank @demariadaniel

demariadaniel (approved):

src/exception/exception-api.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const result = await exceptionService.operations.createProgramException({
+      programId,
+      records,
+    });
+
+    if (!result.successful) {
+      return res.status(422).send(result);
+    }
+    return res.status(201).send(result);
```
joneubank, line 62: A couple things that need fixing in this block.

1. All the code after the try/catch won't run if we throw an error, so we're doing this extra manipulation of a mutable variable `records` for no reason. Just move all this code into the try block.
2. Won't compile, there's a type error. Specificly, `createProgramException` is expecting `records` to be of type `ProgramExceptionRecord` and we only know that they are `ReadonlyArray<TsvUtils.TsvRecordAsJsonObj>`. So we need to do some validation to make sure our records have the correct structure. This actually turns out to be just tricky enough that I spent 30 minutes writing out a solution (see next comment).

joneubank: First, we need some type validaters with user defined type guards: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

For ProgramExceptionRecord this might look like:
```TypeScript
export const isProgramExceptionRecord = (input: any): input is ProgramExceptionRecord => {
  return (
    // input must not be null and be an object (typeof null = 'object', amusingly)
    typeof input === 'object' &&
    input !== null &&
    // program_name must exist and be string
    'program_name' in input &&
    typeof input.program_name === 'string' &&
    // schema must exist and be string
    'schema' in input &&
    typeof input.schema === 'string' &&
    // requested_core_field must exist and be string
    'requested_core_field' in input &&
    typeof input.requested_core_field === 'string' &&
    // requested_exception_value must exist and be string and be in enum list
    'requested_exception_value' in input &&
    typeof input.requested_core_field === 'string' &&
    Object.values(ExceptionValue).includes(input.requested_core_field)
  );
};
```

This can validate a single item of any type matches our interface, and tell TypeScript that it is safe to treat it as such.

This doesn't work for the whole array though. In order to type check the whole array we need another tool:

```TypeScript
export const isArrayOf = <T>(input: any[], validator: (_: any) => _ is T): input is T[] => {
  return input.every(validator);
};
export const isReadonlyArrayOf = <T>(input: ReadonlyArray<any>, validator: (_: any) => _ is T): input is ReadonlyArray<T> => {
  return input.every(validator);
};
```

One version for a standard array, and one that can handle the ReadonlyArray type (couldn't combine nicely :( )

joneubank: Finally, we can use these in the `createProgramException` request handler:
```TypeScript
try {
      const records = await TsvUtils.tsvToJson(file.path);
      if (records.length === 0) {
        throw new Error('TSV has no records!');
      }
      if (!isReadonlyArrayOf(records, isProgramExceptionRecord)) {
        throw new Error('TSV is incorrectly structured');
      }
      const result = await exceptionService.operations.createProgramException({
        programId,
        records,
      });

      if (!result.successful) {
        return res.status(422).send(result);
      }
      return res.status(201).send(result);
    } catch (err) {
      L.error(`Program Exception TSV_PARSING_FAILED`, err);
      return ControllerUtils.unableToProcess(res, ProgramExceptionErrorMessage.TSV_PARSING_FAILED);
    }
  }
  ```

File: src/exception/exception-api.ts [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+const isValidCreateBody = (req: Request, res: Response): boolean => {
```
joneubank, line 66: This is specifically checking for file only, maybe it should be named `requestContainsFile` to be clear.

src/exception/exception-api.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class ExceptionController {
+  @HasFullWriteAccess()
+  async createProgramException(req: Request, res: Response) {
+    if (!isValidCreateBody(req, res)) {
+      return false;
+    }
+
+    const programId = req.params.programId;
```
joneubank, line 39: `programId` has type `any`. We need to validate that `req.params.programId` was provided and is a string.
```suggestion
    const programId = req.params.programId;
    if (!programId && programId === 'string') {
      ControllerUtils.unableToProcess(res, `Invalid ProgramID provided`);
    }
```

src/exception/exception-api.ts
joneubank: You'll also need to update the `records` argument type in a couple of your service methods to be `ReadonlyArray<ProgramExceptionRecord>` instead of `ProgramExceptionRecord[]`

src/utils.ts
```diff
@@ -121,6 +121,10 @@ export namespace ControllerUtils {
     return res.status(503).send({ message });
   };
 
+  export const unableToProcess = (res: Response, message: string): any => {
+    return res.status(422).send({ message });
```
joneubank, line 125: Don't think I've ever seen a 422 in the wild, is kinda cool. According to what I read though 422 implies the user semantics are correct (ie. correctly formed file). I think we just want a 403 response here.

joneubank (changes requested):
Some types feedback for validating the incoming request contents. The validations in the service module look strong, nice work.

src/exception/exception-api.ts
ciaranschutte: the programId comes from the endpoint url, so no programId is just not an endpoint
I will check if express has all params as string

src/exception/exception-api.ts
ciaranschutte: Very cool. thanks for typing this up. this then does the job of most of the validation in exception-service?

src/exception/exception-api.ts
joneubank: oh fair enough. if its a path param then it has to be a string. i'm fine with that.

src/exception/exception-api.ts
joneubank: Maybe make a new file just for the validations, keep it separate from the application logic.

joneubank (approved):
Awesome, thanks!

PR #833, opened 2022-11-01
Bug/2417 submitted data surgery entities

demariadaniel:
**Description of changes**

- Retrieves Donor Clinical Data where Treatment Type is Surgery

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/common-model/functions.ts
+++ b/src/common-model/functions.ts
@@ -70,7 +70,10 @@ export function getClinicalObjectsFromDonor(
     }
   }
 
-  if (clinicalEntitySchemaName === ClinicalEntitySchemaNames.TREATMENT) {
+  if (
+    clinicalEntitySchemaName === ClinicalEntitySchemaNames.TREATMENT ||
+    clinicalEntitySchemaName === ClinicalEntitySchemaNames.SURGERY
+  ) {
     if (donor.treatments) {
       return donor.treatments;
     }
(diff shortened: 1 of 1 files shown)


ciaranschutte:
overall looks great - i'm not 100% sure how well defined the data coming into the function is, so a couple more checks might be beneficial. I'll comment and you can decided for yourself.

we should probably steer clear of using `any` type

src/common-model/functions.ts
```diff
@@ -134,39 +137,64 @@ export function getClinicalEntitySubmittedData(
...
+    }
+    case ClinicalEntitySchemaNames.TREATMENT: {
+      clinicalRecords = result.map((treatment: any) => {
+        const clinicalInfo = treatment.clinicalInfo || {};
+        const therapy_type =
+          treatment.therapies.length === 1
+            ? { therapy_type: treatment.therapies[0].therapyType }
+            : {};
+        const therapy_info =
+          treatment.therapies.length === 1 && treatment.therapies[0].clinicalInfo;
```
ciaranschutte, line 159: checking for `treatment.therapies.length` and dipping into array twice here.
 
 - Will `treatment.therapies` at least be an empty array?
 - cleaner way to do it? maybe something like...
 ```
 const therapy = treatment.therapies && treatment.therapies.length > 0 ?  treatment.therapies[0] : {} ;
 const {therapyTypes = { }, clinicalInfo = {}} = therapy;
 
 ```

src/common-model/functions.ts
```diff
@@ -70,7 +70,10 @@ export function getClinicalObjectsFromDonor(
     }
   }
 
-  if (clinicalEntitySchemaName === ClinicalEntitySchemaNames.TREATMENT) {
+  if (
+    clinicalEntitySchemaName === ClinicalEntitySchemaNames.TREATMENT ||
+    clinicalEntitySchemaName === ClinicalEntitySchemaNames.SURGERY
+  ) {
```
ciaranschutte, line 76: these nested `ifs`.... not sure if there's a reason for them but there's a lot here already

ciaranschutte: can put entire condition in a single if

src/common-model/functions.ts
```diff
@@ -134,39 +137,64 @@ export function getClinicalEntitySubmittedData(
...
+          ...therapy_type,
+          ...therapy_info,
+        };
+      });
+      break;
+    }
+    case ClinicalEntitySchemaNames.SURGERY: {
+      clinicalRecords = result
+        .filter(therapy =>
+          therapy.clinicalInfo.treatment_type.find((type: string) => type.match(/surgery/i)),
```
ciaranschutte, line 174: `therapy.clinicalInfo.treatment_type` can this be undefined potentially?

src/common-model/functions.ts
joneubank: At first I thought this was just about combining the two if statements, like follows
```TypeScript
[ClinicalEntitySchemaNames.TREATMENT, ClinicalEntitySchemaNames.SURGERY].includes(clinicalEntitySchemaName) && donor.treatments```

But then i looked at the rest of the function... definitely fine to continue this pattern as is.

This probably should've been written as a switch:
```TypeScript
switch(clinicalEntitySchemaName( {
  case ClinicalEntitySchemaNames.FAMILY_HISTORY:
   return donor.familyHistory || [];
  case ClinicalEntitySchemaNames.TREATMENT:
    // fall through to SURGERY
  case ClinicalEntitySchemaNames.SURGERY:
   return donor.treaments || [];
   // ... etc for each SchemaName
  default:
    return [];
```
But even that is messy because of the alternate return type.

joneubank: Actual review comment:
In the case of `clinicalEntitySchemaName === SURGERY` I think we only want to return the surgery treatments. The code on line 106 handles that case. Why is surgery so special that it is instead included here?

src/common-model/functions.ts
demariadaniel: Yes, I've added Types now to this switch section + handling for undefined, thanks!

src/common-model/functions.ts
demariadaniel: @joneubank that is a very valid question however there is definitely a bug here, although I think the local Mongo data I'm using may be the issue

I can find several examples of `donor.treatments > treatment.clinicalInfo.therapy_type === 'Surgery'` but cannot find any data (using local Mongodb collection) where `donor.treatments > treatment.therapies[n].therapyType === 'surgery'`

Therapies makes the most sense and I did notice a mismatch btwn data in Clinical Info and the surgery schema fields which suggested something was wrong

But w/ no data representing a case where `therapies[n].therapyType === 'surgery'` I totally overlooked the block on line 106. I had to return `donor.treatments` to even _find_ 'Surgery' data

demariadaniel: @joneubank on further inspection there's actually a working example of this page in Dev > DATA-CA, I think the issue needs to be better defined to specific env/program where the bug is occurring to investigate further

src/common-model/functions.ts
demariadaniel: I've added Types here + yes treatment.therapies should always be an array

demariadaniel: Agree w/ feedback but need to give this a more thorough review b/c it relates to `Therapy` and the ticket is focused on `Surgery` bug, need to limit to refactoring existing logic
I'm questioning if/how this logic should be changed but that's out of scope to ticket

ciaranschutte:
@demariadaniel code looks ok. i think tests should pass once you update with `develop` branch changes

src/common-model/functions.ts
ciaranschutte: Just a cleaner (imho) way of writing the same logic that's been refactored so the same conditional isn't being checked and the same array item is being pulled out. No worries if you want to leave it

src/common-model/functions.ts
```diff
@@ -134,39 +128,66 @@ export function getClinicalEntitySubmittedData(
...
+          ...therapy_info,
+        };
+      });
+      break;
+    }
+    case ClinicalEntitySchemaNames.SURGERY: {
+      clinicalRecords = result
+        .filter((treatment: Treatment) => {
+          const clinicalInfo: ClinicalInfo = treatment.clinicalInfo;
+          const treatmentType = <string[]>clinicalInfo['treatment_type'];
```
ciaranschutte, line 166: for consistency the project as most type assertions in the form of `as` rather than angle brackets.
ie. `const treatmentType = clinicalInfo['treatment_type'] as string[];

PR #837, opened 2022-11-09
Bug/544 fix tsv endpoint bugs

demariadaniel:
**Description of changes**

- Updates Swagger file to correctly append multiple values in URL (entityType, donorId, submitterId)
- Make entityType required + return errorResponse for empty value (i.e. request must specify file(s) to download)
- Better entityType validation (match against schemaNames / entity Alias names i.e. sample_registration or sampleRegistration)

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/clinical/clinical-api.ts
+++ b/src/clinical/clinical-api.ts
@@ -116,26 +116,26 @@ class ClinicalController {
     const programId: string = req.params.programId;
     if (!programId) {
       return ControllerUtils.badRequest(res, 'Invalid programId provided');
+    } else if (!req.query.entityTypes) {
+      return ControllerUtils.badRequest(res, 'Must request specific EntityTypes');
     }
     const sort: string = 'donorId';
     const page: number = 0;
     const state: CompletionStates = req.query.completionState || CompletionStates.all;
     const completionState: {} = completionFilters[state] || {};
 
+    const schemaNames = Object.values(ClinicalEntitySchemaNames);
     const entityTypes: string[] =
-      req.query.entityTypes !== undefined
-        ? req.query.entityTypes
-            .split(',')
-            .map((entityName: EntityAlias & ClinicalEntitySchemaNames) => {
-              const schemaNames = Object.values(ClinicalEntitySchemaNames);
-              if (queryEntityNames.includes(entityName)) {
-                return entityName;
-              } else if (schemaNames.includes(entityName)) {
-                return aliasEntityNames[entityName];
-              } else return undefined;
-            })
-            .filter(Boolean)
-        : [];
+      req.query.entityTypes
+        .split(',')
+        .map((entityName: EntityAlias & ClinicalEntitySchemaNames) => {
+          if (queryEntityNames.includes(entityName)) {
+            return entityName;
+          } else if (schemaNames.includes(entityName)) {
+            return aliasEntityNames[entityName];
+          } else return undefined;
+        })
+        .filter(Boolean) || [];
 
     const donorIds: number[] = parseDonorIdList(req.query.donorIds) || [];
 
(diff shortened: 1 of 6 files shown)


src/common-model/entities.ts
```diff
@@ -42,6 +42,7 @@ export enum ClinicalEntitySchemaNames {
 export type EntityAlias =
   | 'donor'
   | 'sampleRegistration'
+  | 'specimen'
   | 'specimens'
```
ciaranschutte, line 46: this looks off to me. Shouldn't it just be either 'specimen' or 'specimens'?
looking at current dictionary I think it should be 'specimen' as you've added but then can we safely remove 'specimens'?

src/clinical/clinical-api.ts
```diff
@@ -111,14 +116,26 @@ class ClinicalController {
...
+      req.query.entityTypes
+        .split(',')
+        .map((entityName: EntityAlias & ClinicalEntitySchemaNames) => {
+          if (queryEntityNames.includes(entityName)) {
+            return entityName;
+          } else if (schemaNames.includes(entityName)) {
+            return aliasEntityNames[entityName];
+          } else return undefined;
+        })
+        .filter(Boolean) || [];
```
ciaranschutte, line 138: believe we get an array from filter either way so no need for `|| []`
no passing conditions will be an empty array

src/common-model/entities.ts
```diff
@@ -81,6 +82,8 @@ export const aliasEntityNames: Record<ClinicalEntitySchemaNames, EntityAlias> =
   biomarker: 'biomarker',
 };
 
+export const queryEntityNames = <EntityAlias[]>Object.values(aliasEntityNames);
```
ciaranschutte, line 85: can we keep the casting 'as' syntax over this prefix syntax for consistency?

src/resources/swagger.yaml
```diff
@@ -996,20 +996,24 @@ paths:
         - $ref: '#/components/parameters/PathProgramId'
         - in: query
           name: entityTypes
+          explode: false
+          required: true
```
ciaranschutte, line 1000: nice!

ciaranschutte (changes requested):

ciaranschutte:
@demariadaniel merged in the version mergeback and daniels swagger fix

src/common-model/entities.ts
demariadaniel: It is off. The Dictionary defines it as `specimen` but our Donor Model defines it as `specimens`. 
Same issue with `treatment`/`treatments` and several other entities where we have two different sets of aliases.

So I have a lot of code to handle both sets of keys, but it may be redundant in some places and could likely use it's own review.

src/clinical/clinical-api.ts
demariadaniel: I thought so, was just being cautious knowing it might come up in review ;) will remove

src/resources/swagger.yaml
demariadaniel: Can't believe there's a genuine `explode` keyword

src/common-model/entities.ts
demariadaniel: Sure

src/common-model/entities.ts
demariadaniel: OK w/ the updated validation logic I can remove `specimen` b/c we are checking against schema names

There are other places in the Submitted Data code which should be updated to accept either schema name || donor model aliases and that's what I was trying to address, but that should be dine differently and in another ticket

ciaranschutte: awesome thanks for looking into that

PR #842, opened 2022-11-10
Bug 2417 Submitted Data - Missing Surgery Entities

demariadaniel:
**Description of changes**

- Only uses pagination limits when querying `['donor']` else we query ALL Donor records to find entity records

**Type of Change**

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/clinical/donor-repo.ts
+++ b/src/clinical/donor-repo.ts
@@ -217,7 +217,7 @@ export const donorDao: DonorRepository = {
     // Specific Requests for Donor documents can be paginated at the MongoDB level
 
     const limit =
-      entityTypes === ['donor'] && pageSize
+      entityTypes.length === 1 && entityTypes.includes('donor') && pageSize
         ? pageSize
         : await DonorModel.countDocuments({ programId });
 
--- a/src/clinical/donor-repo.ts
+++ b/src/clinical/donor-repo.ts
@@ -217,7 +217,7 @@ export const donorDao: DonorRepository = {
     // Specific Requests for Donor documents can be paginated at the MongoDB level
 
     const limit =
-      entityTypes.includes('donor') && pageSize
+      entityTypes === ['donor'] && pageSize
         ? pageSize
         : await DonorModel.countDocuments({ programId });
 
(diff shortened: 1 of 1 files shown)


src/clinical/donor-repo.ts
```diff
@@ -258,7 +258,7 @@ export const donorDao: DonorRepository = {
     );
 
     const { totalDocs: totalDonors } = result;
-
+    console.log('totalDonors', totalDonors);
```
ciaranschutte, line 261: console.log :)

src/clinical/donor-repo.ts
```diff
@@ -217,7 +217,7 @@ export const donorDao: DonorRepository = {
     // Specific Requests for Donor documents can be paginated at the MongoDB level
 
     const limit =
-      entityTypes.includes('donor') && pageSize
+      entityTypes === ['donor'] && pageSize
```
ciaranschutte, line 220: This will be comparing by "reference". I took a look and don't understand the logic here.

Is the goal to check if `entityTypes` _only_ has `donor`?

ciaranschutte (changes requested):

src/clinical/donor-repo.ts
demariadaniel: You're right we are checking if entityTypes only includes donor
There's likely a better way that isn't checking by reference
Basically we are only concerned with pageSize at this stage if we are querying Donors specifically
All other entities we need to query all results, else we don't know the total number of records (b/c entities are stored as properties of Donors)

src/clinical/donor-repo.ts
demariadaniel: Removed!

src/clinical/donor-repo.ts
ciaranschutte: Gotcha. Sorry I should clarify. 
``` javascript
['donor'] === ['donor']; // this will always be false because they're different objects
```
To check if entityTypes only includes donor, something like this should work:
``` javascript
entityTypes.length === 1 && entityTypes.includes('donor');
```

demariadaniel: Updated

ciaranschutte (approved):

PR #890, opened 2023-02-04
Feat/#880_ GET and DELETE improved error handling

ciaranschutte:
**Description of changes**
- stop throwing errors all the way up the stack, do explicit falsey check for returned documents. no docs found will not throw an error but return [] or null depending on query.
- better error handling, responses and messages
- add middle layer for GET and DELETE requests for uniformity and cleaner distinction of responsibilities (error message comes from "service" file)
- return lean mongo document directly from api instead of using our util
- use `undefined` consistently instead of `nulls`
- adds correct http responses

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

**Checklist before requesting review:**

- [ ] Check branch (code change PRs go to `develop` not master)
- [ ] Check copyrights for new files
- [ ] Manual testing
- [ ] Regression tests completed and passing (double check number of tests).
- [ ] Spelling has been checked.
- [ ] Updated swagger docs accordingly (check it's still valid)
- [ ] Set `validationDependency` in meta tag for [Argo Dictionary](https://github.com/icgc-argo/argo-dictionary) fields used in code

PR #892, opened 2023-02-04
remove no null rule

ciaranschutte:
**Description of changes**

currently only "undefined" is allowed, not "null"

I think "null" should be allowed
this is the rule i'm suggesting we remove https://palantir.github.io/tslint/rules/no-null-keyword/

use case: https://github.com/icgc-argo/argo-clinical/pull/890/files#diff-b3e63aecd11562b14d8eafc010c5aa655a6818a2ae737260971fc881e6289be4R43

returning nulls for db queries  and returning nulls for api responses, eg. `error: null` for no error, `error: undefined` implies it's left to be defined.

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -36,7 +36,6 @@
     ],
     "no-internal-module": true,
     "no-trailing-whitespace": true,
-    "no-null-keyword": true,
     "prefer-const": true,
     "jsdoc-format": true
   }


caravinci:
Interesting proposal, and I'm curious to hear why returning null could be better than returning undefined, or even returning an "empty" array/object for example.
Perhaps undefined may give the wrong impression of there being something wrong with the logic? I can see an issue with that for sure, and thus I avoid doing that too. 

For context, semantically, I've always seen undefined as "that hasn't been defined yet (but could be in the future)"; while null has always meant to me "it is defined, and doesn't exist (and likely won't ever)". The nothingness of it seems absolute, synonimous to zero in math.
Zero doesn't become one or ten, it's always zero, so it's a meaningful value.

Furthermore, though both null and undefined are perfectly valid choices, I tend to use null very sparingly, for it sort of breaks JavaScript's function argument "default value" setting (which relies on the input being undefined).
e.g. in function (str = 'stuff') {} I want to have that fallback "stuff", but null is taken as a "valid" str, which requires extra code guards within the function. TypeScript helps but it feels uneasy to me relying exclusively on a library for safety.

That said, personally, in the case of empty function returns or server responses, I prefer [] or {} or '' or whatever depending on response value type, over null. More declarative, and less prone to breaking other functions relying on that output (for the reason mentioned above.)

These are my thoughts on this subject, but I of course am aware I don't know it all, and I don't what I don't know, so I'm truly interested to understand your motivation a bit better. May I bug you to please elaborate on it? @ciaranschutte 👍

ciaranschutte:
> .. Perhaps undefined may give the wrong impression of there being something wrong with the logic? I can see an issue with that for sure, and thus I avoid doing that too.

The impression of something being wrong with the logic is a consideration yes.

> Furthermore, though both null and undefined are perfectly valid choices, I tend to use null very sparingly, for it sort of breaks JavaScript's function argument "default value" setting (which relies on the input being undefined). e.g. in function (str = 'stuff') {} I want to have that fallback "stuff", but null is taken as a "valid" str, which requires extra code guards within the function. TypeScript helps but it feels uneasy to me relying exclusively on a library for safety.

^ this is interesting and a good use case against

> 
> That said, personally, in the case of empty function returns or server responses, I prefer [] or {} or '' or whatever depending on response value type, over null. More declarative, and less prone to breaking other functions relying on that output (for the reason mentioned above.)

Motivation is being explicit about the lack of value and to me it flows better than `undefined`.

Specific case: a GET query for a resource
mongoose will return `null` if the resource is not found and from there I pass that through to my api responses
```
{
    "exception": null,
    "error": "Cannot find program '-xx",
    "validationErrors": [],
    "success": false
}
```
or for a happy result:
```
{
    "exception": Exception,
    "error": null
    "validationErrors": [],
    "success": true
}
```
then again most of the code base uses `undefined`

On further reflection I'm going to stick with `undefined` and leave the rule as is.
It's worth more to keep the consistency in this case I think.

PR #909, opened 2023-02-24
🐛 #2580 Update Submitted Data Specimen Rules

demariadaniel:
**Description of changes**

Updates Submitted Data Core Completion record rules
- If CoreCompletion.Specimens > 0, find ratio of Submitted Specimens / Registered Samples

Refactors Submitted Data + helper functions to clarify code

**Type of Change**

- [x] Bug
- [x] Refactor
- [ ] New Feature
- [ ] Release Candidate

**Checklist before requesting review:**

- [ ] Check branch (code change PRs go to `develop` not master)
- [ ] Check copyrights for new files
- [ ] Manual testing
- [ ] Regression tests completed and passing (double check number of tests).
- [ ] Spelling has been checked.
- [ ] Updated swagger docs accordingly (check it's still valid)
- [ ] Set `validationDependency` in meta tag for [Argo Dictionary](https://github.com/icgc-argo/argo-dictionary) fields used in code

PR #910, opened 2023-02-24
🧹 Submitted data sort refactor

demariadaniel:
**Description of changes**

Cleans up Sorting functions  

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/clinical/service-worker-thread/tasks.ts
+++ b/src/clinical/service-worker-thread/tasks.ts
@@ -107,11 +107,11 @@ const sortDocs = (sort: string, entityName: string, completionStats: CompletionR
 
   // -1 Current lower index than Next, +1 Current higher index than Next
   let order = 0;
-  const isDescending = sort.includes('-') ? -1 : 1;
+  const isDescending = sort.startsWith('-') ? -1 : 1;
 
   if (isDefaultSort) {
-    const { donorId: currentDonorId } = currentRecord as Donor;
-    const { donorId: nextDonor } = nextRecord as Donor;
+    const { donorId: currentDonorId } = currentRecord;
+    const { donorId: nextDonor } = nextRecord;
 
     // Sort Clinically Incomplete donors to top (sorted by donorId at DB level)
     const completionA =
@@ -134,13 +134,7 @@ const sortDocs = (sort: string, entityName: string, completionStats: CompletionR
     const first: ClinicalInfo[string] = currentEntry[key] !== undefined && currentEntry[key];
     const next: ClinicalInfo[string] = nextEntry[key] !== undefined && nextEntry[key];
 
-    let valueSort: number;
-
-    if (first && next) {
-      valueSort = first > next ? 1 : first === next ? 0 : -1;
-    } else {
-      valueSort = first && !next ? 1 : next ? -1 : 0;
-    }
+    const valueSort = first === next ? 0 : first && (!next || first > next) ? 1 : -1;
 
     order = valueSort;
   }
--- a/src/clinical/service-worker-thread/tasks.ts
+++ b/src/clinical/service-worker-thread/tasks.ts
@@ -159,6 +159,7 @@ const sortRecordsByColumn = (
   return valueSort;
 };
 
+// Formats + Organizes Clinical Data
 const mapEntityDocuments = (
   entity: EntityClinicalInfo,
   originalResultsArray: EntityClinicalInfo[],
@@ -216,7 +217,8 @@ const mapEntityDocuments = (
   };
 };
 
-export function filterDonorIdDataFromSearch(donors: Donor[], query: ClinicalSearchQuery) {
+// Submitted Data Search Results
+function FilterDonorIdDataFromSearch(donors: Donor[], query: ClinicalSearchQuery) {
   const { donorIds, submitterDonorIds } = query;
 
   const useFilteredDonors =
@@ -280,7 +282,7 @@ function extractDataFromDonors(donors: Donor[], schemasWithFields: any) {
 }
 
 // Main Clinical Entity Submitted Data Function
-export function extractEntityDataFromDonors(
+function extractEntityDataFromDonors(
   donors: Donor[],
   totalDonors: number,
   schemasWithFields: any,
@@ -343,9 +345,11 @@ export function extractEntityDataFromDonors(
 export enum WorkerTasks {
   ExtractDataFromDonors,
   ExtractEntityDataFromDonors,
+  FilterDonorIdDataFromSearch,
 }
 
 export const WorkerTasksMap: Record<WorkerTasks, Function> = {
   [WorkerTasks.ExtractDataFromDonors]: extractDataFromDonors,
   [WorkerTasks.ExtractEntityDataFromDonors]: extractEntityDataFromDonors,
+  [WorkerTasks.FilterDonorIdDataFromSearch]: FilterDonorIdDataFromSearch,
 };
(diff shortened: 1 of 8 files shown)


ciaranschutte:
looking at the context of where this code lives, i'm wondering if `filterDonorIdDataFromSearch` should be run as a service task using `await runTaskInWorkerThread(task_function)`

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -213,6 +216,35 @@ const mapEntityDocuments = (
   };
 };
 
+export function filterDonorIdDataFromSearch(donors: Donor[], query: ClinicalSearchQuery) {
+  const { donorIds, submitterDonorIds } = query;
+
+  const useFilteredDonors =
+    (donorIds && donorIds.length) || (submitterDonorIds && submitterDonorIds.length);
```
ciaranschutte, line 223: checking using a util function like `isEmpty` makes this line cleaner. right now `useFilteredDonors` evaluates to a `number` type rather than a `boolean`

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -213,6 +216,35 @@ const mapEntityDocuments = (
...
+    ? donors.filter(donor => {
+        const { donorId, submitterId } = donor;
+        const stringId = `${donorId}`;
+        const donorMatch = donorIds?.filter(id => stringId.includes(id));
+        const submitterMatch = submitterId
+          ? submitterDonorIds?.filter(id => submitterId.includes(id))
+          : [];
+        return donorMatch.length > 0 || submitterMatch.length > 0;
+      })
+    : donors;
```
ciaranschutte, line 235: based on the typing of `ClinicalSearchQuery` both `donorIds` and `submitterDonorIds` will always at least be an empty array, so once this is an accurate reflection (eg. default value is accounted for) , there's no need for the optional chaining as it's already an `[]` and not potentially `undefined`

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -213,6 +216,35 @@ const mapEntityDocuments = (
...
+
+  const filteredDonors = useFilteredDonors
+    ? donors.filter(donor => {
+        const { donorId, submitterId } = donor;
+        const stringId = `${donorId}`;
+        const donorMatch = donorIds?.filter(id => stringId.includes(id));
+        const submitterMatch = submitterId
+          ? submitterDonorIds?.filter(id => submitterId.includes(id))
+          : [];
+        return donorMatch.length > 0 || submitterMatch.length > 0;
```
ciaranschutte, line 233: can likely do an early return if `donorMatch` finds what you're looking for, instead of both filter functions.
looks like it's just a check for a single match - can use `Array.some()` maybe so the whole array doesn't have to be iterated?

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -213,6 +216,35 @@ const mapEntityDocuments = (
...
+      })
+    : donors;
+
+  const totalResults = filteredDonors.length;
+  const searchResults = filteredDonors.map((donor: Donor) => {
+    const { donorId, submitterId } = donor;
+    const submitterDonorId = submitterId;
+    return { donorId, submitterDonorId };
```
ciaranschutte, line 241:
```
  const searchResults = filteredDonors.map((donor: Donor) => {
    const { donorId, submitterId } = donor;
    return { donorId, submitterDonorId: submitterId };
}
```

src/common-model/functions.ts
```diff
@@ -172,23 +172,35 @@ export function getClinicalEntitySubmittedData(
   return clinicalRecords;
 }
 
+export const DonorCompletionFields: Array<keyof Donor | EntityAlias> = [
```
ciaranschutte, line 175: `donorCompletionFields` it's just an array like `requiredFields` good to keep naming consistent

src/common-model/functions.ts
```diff
@@ -172,23 +172,35 @@ export function getClinicalEntitySubmittedData(
...
+  'completionStats',
+  'specimens',
+  'followUps',
+  'treatments',
+  'primaryDiagnoses',
+  ClinicalEntitySchemaNames.REGISTRATION,
```
ciaranschutte, line 181: it feels odd to me to reference one value of a different enum here. particularly because these values are camelCased and `ClinicalEntitySchemaNames` are snake cased.

src/clinical/service-worker-thread/tasks.ts
ciaranschutte: looks like they may be undefined coming from frontend though https://github.com/icgc-argo/platform-ui/blob/develop/generated/gql_types.tsx#L163

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -106,18 +106,21 @@ const sortDocs = (sortQuery: string, entityName: string, completionStats: Comple
...
+  const isDescending = sortQuery.startsWith('-');
   const isDefaultSort =
-    entityName === ClinicalEntitySchemaNames.DONOR && sortQuery.includes('donorId');
-  const isDescending = sortQuery.startsWith('-') ? -1 : 1;
-  const key = isDescending === -1 ? sortQuery.split('-')[1] : sortQuery;
+    entityName === ClinicalEntitySchemaNames.DONOR &&
+    sortQuery.includes('completionStats.coreCompletionPercentage');
+
+  const queryKey = isDescending ? sortQuery.split('-')[1] : sortQuery;
+  const key = queryKey === 'donorId' ? 'donor_id' : queryKey;
```
ciaranschutte, line 115: this looks good - wondering where the disconnect is between our naming conventions - something to keep an eye on

ciaranschutte (approved):
few comments that I'll leave to you to take as you see fit

src/clinical/service-worker-thread/tasks.ts
demariadaniel: Went one further

`  const searchResults = filteredDonors.map(({ donorId, submitterId }: Donor) => ({
    donorId,
    submitterDonorId: submitterId,
  }));`

src/common-model/functions.ts
demariadaniel: Yes, I looked into how I can smooth this out yesterday, it's a side effect of sample registrations / specimen record core completion being intertwined, working on it

src/clinical/service-worker-thread/tasks.ts
demariadaniel: As far as I can remember and from what I can see when reviewing the codebase, some of this is actually in response to requirements on the UI. This was an issue at the very beginning of building Submitted Data features and is something I was hoping to revisit one day.

ciaranschutte: I suspected it was UI. Is it a case of enforcing consistency with how we name vars in our queries?

src/common-model/functions.ts
demariadaniel: OK I've updated this. `donorCompletionFields` now only includes fields found on Donor core completion. 

`extractEntityDataFromDonors` handling for Sample Registration is also updated.
This could probably be improved a little further for consistency, by moving this handling into `getClinicalEntitySubmittedData`, where all other entities are handled. I'll update this as part of #909

ciaranschutte: sounds good thank you

src/clinical/service-worker-thread/tasks.ts
demariadaniel: There are Types (such as DonorCentricRecord) where it's `donor_id`, but our Donor model uses `donorId`. 
Because the early work on this feature was largely based off of existing code from other features, naming collisions may have started there. I could definitely clean this up a little with some dedicated time to review.

PR #913, opened 2023-03-07
Feat/#843 specimen exception endpoint

ciaranschutte:
**Description of changes**

- mostly moving functions and types into common modules / middleware for use in both program and entity exceptions
- moves parseTSV and upload to middlewares
- adds api, service, and repo calls for entity exception endpoints
- cleans up typing for clarity and reusability

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/exception/exception-api.ts
+++ b/src/exception/exception-api.ts
@@ -21,7 +21,7 @@ import { NextFunction, Request, Response } from 'express';
 import { HasFullReadAccess, HasFullWriteAccess } from '../decorators';
 import { loggerFor } from '../logger';
 import { ControllerUtils, TsvUtils } from '../utils';
-import { RepoError } from './exception-repo';
+import { RepoError } from './repo/types';
 import * as exceptionService from './exception-service';
 import { isReadonlyArrayOf } from './types';
 
@@ -71,12 +71,26 @@ class ExceptionController {
     return res.status(getResStatus(result)).send(result);
   }
 
-  // donor level exceptions
-  async createDonorException(req: Request, res: Response) {
-    // check if program exception exists
-    // const programExceptionExists = await exceptioncheckProgramException();
+  @HasFullWriteAccess()
+  async createEntityException(req: Request, res: Response) {
+    const programId = req.params.programId;
+    const existingProgramException = await exceptionService.operations.getProgramException({
+      programId,
+    });
 
-    return res.status(400).send({});
+    if (existingProgramException.exception !== undefined) {
+      L.debug('program exception exists already');
+      return res.status(400).send('program exception already exists');
+    }
+
+    const records = res.locals.records;
+    const result = await exceptionService.operations.createEntityException({
+      programId,
+      records,
+    });
+
+    const status = !result.success ? 422 : 201;
+    return res.status(status).send(result);
   }
 }
 
--- a/src/exception/exception-api.ts
+++ b/src/exception/exception-api.ts
@@ -23,7 +23,7 @@ import { loggerFor } from '../logger';
 import { ControllerUtils, TsvUtils } from '../utils';
 import { RepoError } from './repo/types';
 import * as exceptionService from './exception-service';
-import { isReadonlyArrayOf } from './types';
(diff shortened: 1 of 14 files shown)


src/exception/types.ts
```diff
@@ -65,6 +87,8 @@ export const isProgramExceptionRecord = (input: any): input is ProgramExceptionR
   );
 };
 
+export const isEntityExceptionRecord = (input: any): input is any => true;
```
ciaranschutte, line 90: this is extra validation

ciaranschutte:
mostly moving functions and types into common modules / middleware for use in both program and entity exceptions

src/exception/types.ts
ciaranschutte: leaving as is to not break parseTsv func contract

joneubank:
Can you add description to the PR with summary of the work performed?

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const update = { $set: { programId: exception.programId, specimen: exception.specimen } };
+
+    try {
+      return await EntityExceptionModel.findOneAndUpdate(
+        { programId: exception.programId },
+        update,
+        { upsert: true, new: true },
+      ).lean(true);
+      // L.info(`doc created ${doc}`);
```
joneubank, line 62: Remove unused commented code.

src/exception/types.ts
joneubank: Does this do anything at all? This looks like placeholder code for actual implementation, if so there should at least be some comments explaining its purpose.

This actually seems quite dangerous as it will take any variable and make it `any` type, removing all type checking.

src/routes/exception.ts
```diff
@@ -19,19 +19,29 @@
...
-
 const upload = multer({ dest: '/tmp' });
 
+// routes
+router.post('*', upload.single('exceptionFile'), requestContainsFile);
+
 router.post(
   '/',
-  upload.single('programExceptionFile'),
+  parseTSV(isProgramExceptionRecord),
```
joneubank, line 35: Using parseTSV here as middleware instead of in the service I think causes us to lose our type safety... see other comments

joneubank (changes requested):
Comments are mostly about the refactoring to move the parseTSV into express handlers which causes use to lose type checking... can we look at simplifying this so we don't end up passing around thigns typed as `any`?

ciaranschutte:
@joneubank 
- moved parseTSV back into the api function calls, using individual type guards to narrow `records` to specific types. Couldn't get this working with a single function like `(typeguard) =>  (validate as typeguard) => ...`

- updated `parseTsv` to use the global err handler

- rewrote `isEntityExceptionRecord` type guard instead of placeholder function

src/exception/exception-api.ts
```diff
@@ -41,39 +37,37 @@ function getResStatus(result: exceptionService.Result): number {
...
 
+const validateProgramExceptionRecords = (records: ReadonlyArray<TsvUtils.TsvRecordAsJsonObj>) => {
+  if (!isReadonlyArrayOf(records, isProgramExceptionRecord)) {
+    L.debug(`Program Exception TSV_PARSING_FAILED`);
+    throw new Errors.TSVParseError();
+  }
+  return records;
+};
+
+const validateEntityExceptionRecords = (records: ReadonlyArray<TsvUtils.TsvRecordAsJsonObj>) => {
```
joneubank, line 48: To help future devs understand what this is doing, let's explicitly declare the return type:

```suggestion
const validateEntityExceptionRecords = (records: ReadonlyArray<TsvUtils.TsvRecordAsJsonObj>): ReadonlyArray<EntityExceptionRecord> => {
```

joneubank: Same for `validateProgramExceptionRecords` on line 40

joneubank (commented):
Great updates, types are happy again!

joneubank (approved):
Thanks for all the chagnes!

PR #928, opened 2023-03-21
Feat/856 Validate Reference Radiation IDs

demariadaniel:
**Description of changes**

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/argo-clinical/856

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/submission/validation-clinical/therapy.ts
```diff
@@ -45,14 +45,101 @@ export const validate = async (
...
+  checkTreatmentHasCorrectTypeForTherapy(therapyRecord, treatment, errors);
+
+  const {
+    clinicalInfo: { treatment_type },
+    therapies,
+  } = treatment;
+
+  const isRadiationTreatment =
+    Array.isArray(treatment_type) &&
+    treatment_type?.includes('Radiation therapy') &&
```
ciaranschutte, line 61: `treatment_type` will always have `includes` property because its core js array api and the previous line makes sure its an array

src/submission/validation-clinical/therapy.ts
```diff
@@ -45,14 +45,101 @@ export const validate = async (
...
+  const radiationErrors: SubmissionValidationError[] = isRadiationTreatment
+    ? radiation(therapyRecord, treatment)
+    : [];
+
+  errors = [...errors, ...radiationErrors];
   return { errors };
 };
 
-export function checkTreatementHasCorrectTypeForTherapy(
+const radiation = (
```
ciaranschutte, line 72: this function name could be more descriptive. similar to the other funcs eg. "checkThisThingForThisThing"

src/submission/submission-error-messages.ts
```diff
@@ -113,6 +115,10 @@ const ERROR_MESSAGES: { [key: string]: (errorData: any) => string } = {
     `The combination of submitter_donor_id '${data.info.submitter_donor_id}' and submitter_treatment_id '${data.info.submitter_treatment_id}' can only be associated with one surgery_type. Please correct your data submission.`,
   DUPLICATE_SURGERY_WHEN_SPECIMEN_NOT_SUBMITTED: data =>
     `When submitter_specimen_id is not submitted, the combination of [submitter_donor_id = '${data.info.submitter_donor_id}' and submitter_treatment_id = '${data.info.submitter_treatment_id}' ] should only be submitted once in the Surgery schema. Please correct your data submission.`,
+  RADIATION_REFERENCE_ID_CONFLICT: () =>
+    `The submitter_treatment_id submitted in the "reference_radiation_treatment_id" field does not exist.`,
+  RADIATION_THERAPY_TREATMENT_CONFLICT: () =>
+    `The submitter_treatment_id submitted in the "reference_radiation_treatment_id" field is not for radiation treatment.`,
```
ciaranschutte, line 121: might be worth sending back the submitted treatment ids too. i believe its available in the data argument

src/submission/validation-clinical/therapy.ts
ciaranschutte: i misread this. this is checking if `treatment_type` is defined? which previous line should cover

ciaranschutte:
Ah I see this is a WIP. Confused it with the other one that's waiting on a review. Maybe we could do "draft pull requests" for WIP that's up as a pr

src/submission/validation-clinical/therapy.ts
demariadaniel: Yes I've updated to `validateRadiationRecords`

src/submission/validation-clinical/therapy.ts
demariadaniel: True, updated

src/submission/validation-clinical/therapy.ts
demariadaniel: So this, I was following the pattern here:
https://github.com/icgc-argo/argo-clinical/blob/master/src/submission/validation-clinical/index.ts#L36

And thinking we should maybe export `radiation` as it's own collection of validators the way we do with other entities. But maybe I'll just leave it as part of `treatment` for now rather than mess with the existing setup.

src/submission/submission-service.ts
```diff
@@ -356,9 +356,9 @@ export namespace operations {
     command: MultiClinicalSubmissionCommand,
   ): Promise<CreateSubmissionResult> => {
     // get program or create new one
-    let exsistingActiveSubmission = await submissionRepository.findByProgramId(command.programId);
-    if (!exsistingActiveSubmission) {
-      exsistingActiveSubmission = await submissionRepository.create({
+    let existingActiveSubmission = await submissionRepository.findByProgramId(command.programId);
```
demariadaniel, line 359: Updated spelling errors

src/submission/validation-clinical/therapy.ts
```diff
@@ -45,14 +46,122 @@ export const validate = async (
...
+        ...errors,
+        utils.buildSubmissionError(
+          therapyRecord,
+          DataValidationErrors.REFERENCE_RADIATION_ID_CONFLICT,
+          TreatmentFieldsEnum.submitter_treatment_id,
+          {
+            [TreatmentFieldsEnum.treatment_type]: treatment_type,
+            therapyType: ClinicalEntitySchemaNames.RADIATION,
+            reference_radiation_treatment_id,
+          },
```
demariadaniel, line 155: This final validation needs one last pass to get working

src/submission/submission-error-messages.ts
demariadaniel: This is now added!

src/submission/validation-clinical/surgery.ts
```diff
@@ -35,7 +35,7 @@ import {
   DonorFieldsEnum,
   SurgeryFieldsEnum,
 } from '../../common-model/entities';
-import { checkTreatementHasCorrectTypeForTherapy, getTreatment } from './therapy';
+import { checkTreatmentHasCorrectTypeForTherapy, getTreatment } from './therapy';
```
demariadaniel, line 38: More spelling

src/submission/validation-clinical/therapy.ts
demariadaniel: Updated, tested + working

src/submission/validation-clinical/therapy.ts
demariadaniel: OK so this has changed quite a bit
First, radiation function is now `validateRadiationRecords`
Second, we were previously only validating a single record for a single donor at a time, so more important than separating radiation functions was to create a place for cross file validators
See `src/submission/validation-clinical/therapy.ts @ 70`
This should likely be moved out of therapy into another file but started here

ciaranschutte:
is there by any chance a list of files that have changes other spelling fixes? just wondering before I review comb through as I don't want to miss anything thinking its a speller @demariadaniel

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+  const {
+    clinicalInfo: {
+      submitter_donor_id: relatedTreatmentDonorId,
+      submitter_treatment_id,
+      treatment_type,
+    },
+  } = treatment;
+
+  const {
+    submitter_donor_id: submittedTherapyDonorId,
```
ciaranschutte, line 113: is this renaming to provide more context in the file? ie. in the whole system it's `submitter_donor_id` but we are using it here as `submittedTherapyDonorId`?

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+  } = therapyRecord;
+
+  // Compare across other Submissions
+  let submittedTreatments: DeepReadonly<ClinicalInfo>[] = [];
+
+  for (const donorId in submittedRecords) {
+    const submittedTreatmentRecords = [...submittedRecords[donorId].treatment] as ClinicalInfo[];
+    if (submittedTreatmentRecords && submittedTreatmentRecords.length)
+      submittedTreatments = [...submittedTreatments, ...submittedTreatmentRecords];
+  }
```
ciaranschutte, line 125:
``` typescript

  for (const donorId in submittedRecords) {
    const submittedTreatmentRecords = submittedRecords[donorId].treatment
    if (submittedTreatmentRecords?.length)
      submittedTreatments = [...submittedTreatments, ...submittedTreatmentRecords];
  }
  ```
  
  I think this works the same (ignore typings for now).
  `submittedRecord[donorId].treatment` looks like it should be an array already, so the spread new array thing is redundant

File: src/submission/validation-clinical/therapy.ts [hunk 9]
```diff
@@ -17,42 +17,234 @@
...
+    entityTypes: [ClinicalEntitySchemaNames.TREATMENT, ClinicalEntitySchemaNames.RADIATION],
+    page: 0,
+    sort: 'donorId',
+    donorIds: [],
+    submitterDonorIds: [],
+  };
+
+  // Compare across all Treatments
+  const { donors } =
+    (await donorDao.findByPaginatedProgramId(programId, query)) || emptyDonorResult;
```
ciaranschutte, line 138: I think this is not working as expected. likely the `donorDao` is always returning an array of some kind - TS will tell you the type of `{ donors }` is `const donors: readonly DeepReadonly<Donor>[]`

so `emptyDonorResult` is never assigned and not needed

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+
+  const storedTreatments = donors.length
+    ? donors
+        .map(donor => donor.treatments)
+        .filter(treatments => Boolean(treatments))
+        .flat()
+    : [];
```
ciaranschutte, line 145: you can save an iteration of the array by just using a `filter`
most array methods will give you back empty arrays if operating on empty array 
ie. they won't throw errors for empty arrays
`[].filter(x => x?.y?.z).flat() // result is []`


I think this could be simplified to:
```typescript
  const storedTreatments = donors.filter(donor => Boolean(donor?.treatments)).flat();
```

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+            [TreatmentFieldsEnum.treatment_type]: treatment_type,
+            reference_radiation_treatment_id,
+            therapyType: ClinicalEntitySchemaNames.RADIATION,
+          },
+        ),
+      ];
+    }
+
+    // Therapy + Treatment are associated with Radiation
+    const currentTreatmentIsRadiation = treatment_type && treatment_type === 'radiation';
```
ciaranschutte, line 183: ` const currentTreatmentIsRadiation =  treatment_type === 'radiation';`
not doing any property lookup , so no need to check if the object exists first

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+
+    const submittedTreatmentIsRadiation =
+      submissionTreatmentIdMatch && submissionTreatmentIdMatch.treatment_type === 'radiation';
+
+    const storedTreatmentIsRadiation =
+      storedTreatmentIdMatch && storedTreatmentIdMatch.clinicalInfo?.treatment_type === 'radiation';
```
ciaranschutte, line 189: may as well just stick with optional chaining for the full block rather than half old school object check

ciaranschutte (changes requested):
awesome work ! - thank you for all the clean up too! good separation of logic and handling. couple of syntaxy js suggestions. will take a longer look tomorrow to get my head around the business logic. @joneubank would likely be a better eye for that - suggestining 2nd reviewer before merge

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
   if (!therapyRecord || !mergedDonor || !existentDonor) {
     throw new Error("Can't call this function without a registerd donor & therapy record");
   }
 
-  const errors: SubmissionValidationError[] = [];
+  let errors: SubmissionValidationError[] = [];
   const treatment = getTreatment(therapyRecord, mergedDonor, errors);
   if (!treatment) return { errors };
-  checkTreatementHasCorrectTypeForTherapy(therapyRecord, treatment, errors);
+  // TODO: Refactor use of push
```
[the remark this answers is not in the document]
demariadaniel: Great points!

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+) => {
+  const isRadiationRecord = Object.values(RadiationFieldsEnum).some(field =>
+    Object.keys(therapyRecord).includes(field),
+  );
+
+  const radiationErrors: SubmissionValidationError[] = isRadiationRecord
+    ? await validateRadiationRecords(donor, treatment, therapyRecord, submittedRecords)
+    : [];
+
+  const crossFileErrors = [...radiationErrors];
```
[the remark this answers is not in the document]
demariadaniel: The idea here is that it is only equal to `radiationErrors` until we add more validations
I try not to spend much time solving 'future problems' that might not exist, but my next ~3 tickets are all cross-file validation
So this will look more like:
`const crossFileErrors = [...radiationErrors, ...followUpErrors, ...chemotherapyErrors];`
Given that perspective, let me know if there's a better approach

src/submission/validation-clinical/therapy.ts
demariadaniel: I will give a second look, given I used 'as' this is definitely a crutch specifically to manage Types

src/submission/validation-clinical/therapy.ts
demariadaniel: Worth a review, I added this in response to unit test timeout errors, now that I figured out the real solution this may be unneeded

src/submission/validation-clinical/therapy.ts
demariadaniel: Yes. The intent here is to disambiguate between the Submitter Donor ID associated with the Therapy Record we are currently validating, and any of the multiple Submitter Donor IDs we are matching it against

We are checking, does Submitter Donor ID A match Submitter Donor ID B, C, D, etc., so using the key name is not important necessarily, the end result is just a boolean

demariadaniel: There might be potential to just lump all comparison values together, i.e. we have our `submitter_donor_id` which is from the submitted therapy record, then we loop through a new `relatedTreatments[]` and compare if `submitter_donor_id === relatedTreatments[i].submitter_donor_id`, does that make sense?

src/submission/validation-clinical/therapy.ts
demariadaniel: I see, yes, will update

src/submission/validation-clinical/therapy.ts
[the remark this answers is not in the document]
ciaranschutte: Great! Yes this just didn't click for me on reading - makes sense as is!

src/submission/validation-clinical/therapy.ts
ciaranschutte: i don't think the `as` assertion is needed , submittedRecords is typed already so submittedTreatmentRecords should be grand as is

src/submission/validation-clinical/therapy.ts
demariadaniel: Updated

src/submission/validation-clinical/therapy.ts
demariadaniel: Aha I made this work using `sinon.stub.resolves` in the unit test, and reverted this line to `const { donors } = await donorDao.findByPaginatedProgramId(programId, query);`

src/submission/validation-clinical/therapy.ts
```diff
@@ -17,42 +17,234 @@
...
+const crossFileValidator = async (
+  donor: Donor,
+  treatment: DeepReadonly<Treatment>,
+  therapyRecord: DeepReadonly<SubmittedClinicalRecord>,
+  // submitted records needed to validate current submission
+  submittedRecords: DeepReadonly<ClinicalSubmissionRecordsByDonorIdMap>,
+) => {
+  const isRadiationRecord = Object.values(RadiationFieldsEnum).some(field =>
+    Object.keys(therapyRecord).includes(field),
+  );
```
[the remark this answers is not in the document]
demariadaniel: Updated

src/submission/validation-clinical/therapy.ts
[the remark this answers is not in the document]
demariadaniel: Removed

src/submission/validation-clinical/therapy.ts
demariadaniel: Keeping as is for now, can revisit a map solution at a later time

src/submission/validation-clinical/therapy.ts
demariadaniel: Updated

src/submission/validation-clinical/therapy.ts
demariadaniel: Yep looks like this works

ciaranschutte (approved):

PR #935, opened 2023-03-30
Bug/#925 casing

ciaranschutte:
**Description of changes**

- Adds case insensitivity for records with submitted exception value
- updates record with normalized exception value eg. 'unknown' => 'Unknown'

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/submission/submission-service.ts
+++ b/src/submission/submission-service.ts
@@ -896,49 +896,50 @@ export namespace operations {
           .processParallel(command.clinicalType, record, index, schema);
 
         if (schemaResult.validationErrors.length > 0) {
+          let validationErrors: DeepReadonly<SchemaValidationError>[] = [
+            ...schemaResult.validationErrors,
+          ];
 
-          let validationErrors = schemaResult.validationErrors;
           if (FEATURE_SUBMISSION_EXCEPTIONS_ENABLED) {
             // check for program exception
-          const result = await programExceptionRepository.find(command.programId);
+            const result = await programExceptionRepository.find(command.programId);
 
-          let validationErrors: DeepReadonly<SchemaValidationError>[] = [];
+            validationErrors = [];
 
-          if (isProgramException(result)) {
-            schemaResult.validationErrors.forEach(validationError => {
-              const validationErrorFieldName = validationError.fieldName;
-              const recordValue = record[validationErrorFieldName];
+            if (isProgramException(result)) {
+              schemaResult.validationErrors.forEach(validationError => {
+                const validationErrorFieldName = validationError.fieldName;
+                const recordValue = record[validationErrorFieldName];
 
-              // Zero Array type exceptions exist, but recordValue type is string | string[]
-              if (Array.isArray(recordValue)) {
-                validationErrors.push(validationError);
-                return;
-              }
+                // Zero Array type exceptions exist, but recordValue type is string | string[]
+                if (Array.isArray(recordValue)) {
+                  validationErrors.push(validationError);
+                  return;
+                }
 
-              const normalizedRecordValue = normalizeExceptionValue(recordValue);
+                const normalizedRecordValue = normalizeExceptionValue(recordValue);
 
-              const exceptionExists = checkExceptionExists(
-                result.exceptions,
-                validationError,
-                normalizedRecordValue,
-              );
+                const exceptionExists = checkExceptionExists(
+                  result.exceptions,
+                  validationError,
+                  normalizedRecordValue,
+                );
 
-              if (exceptionExists) {
-                // ensure value is normalized exception value
-                const normalizedExceptionRecord = {
-                  ...record,
-                  ...{ [validationErrorFieldName]: normalizedRecordValue },
-                };
-                processedRecord = normalizedExceptionRecord;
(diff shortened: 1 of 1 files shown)


src/submission/submission-service.ts
```diff
@@ -891,23 +895,55 @@ export namespace operations {
...
+          if (isProgramException(result)) {
+            schemaResult.validationErrors.forEach(validationError => {
+              const validationErrorFieldName = validationError.fieldName;
+              const recordValue = record[validationErrorFieldName];
+
+              // Zero Array type exceptions exist, but recordValue type is string | string[]
+              if (Array.isArray(recordValue)) {
+                validationErrors.push(validationError);
+                return;
+              }
```
ciaranschutte, line 912: there are no record field values or exception values that are currently arrays but this accounts for the typing coming from lectern external lib and future consideration.

ciaranschutte:
changed from a cleaner `map/filter` to `accumulator` style in a `forEach` because we're not just filtering anymore.
If an exception exists after the record value is normalized (eg. 'unknown' => 'Unknown'), the record also needs to be updated with the value.

src/submission/submission-service.ts
```diff
@@ -18,9 +18,9 @@
  */
 
 import { entities as dictionaryEntities } from '@overturebio-stack/lectern-client';
-import { DataRecord } from '@overturebio-stack/lectern-client/lib/schema-entities';
+import { SchemaValidationError } from '@overturebio-stack/lectern-client/lib/schema-entities';
 import { DeepReadonly } from 'deep-freeze';
-import _ from 'lodash';
+import _, { isEmpty, upperFirst } from 'lodash';
```
joneubank, line 23: Can we choose between importing all of lodash or just named methods? Serverside there is no problem with importing all

src/submission/submission-service.ts
```diff
@@ -866,12 +866,16 @@ export namespace operations {
...
-      return !exception ? false : exception.exceptionValue === record[validationErrorField];
+      const exception = exceptions.find(
+        exception => exception.coreField === validationError.fieldName,
+      );
+
+      return exception?.exceptionValue === recordValue;
     }
   };
 
+  const normalizeExceptionValue = (value: string) => upperFirst(value.trim().toLowerCase());
```
joneubank, line 877: 👍

src/submission/submission-service.ts
```diff
@@ -852,10 +852,10 @@ export namespace operations {
     );
   }
 
-  const applyExceptionIfExists = (
+  const checkExceptionExists = (
```
joneubank, line 855: This could use a docstring comment. I think the method returns true if the exception value is permissible because of an exception registered for this program?

src/submission/submission-service.ts
```diff
@@ -892,25 +896,59 @@ export namespace operations {
           .processParallel(command.clinicalType, record, index, schema);
 
         if (schemaResult.validationErrors.length > 0) {
-          let validationErrors = schemaResult.validationErrors;
+          let validationErrors: DeepReadonly<SchemaValidationError>[] = [
```
joneubank, line 899: Not an issue, but you don't need to declare the type here since it will be infered from `schemaResult.validationErrors`

src/submission/submission-service.ts
```diff
@@ -892,25 +896,59 @@ export namespace operations {
...
+                  result.exceptions,
+                  validationError,
+                  normalizedRecordValue,
+                );
+
+                if (exceptionExists) {
+                  // ensure value is normalized exception value
+                  const normalizedExceptionRecord = {
+                    ...record,
+                    ...{ [validationErrorFieldName]: normalizedRecordValue },
```
joneubank, line 932: spread + object = don't need either
just put `[validationErrorFieldName]: normalizedRecordValue`

src/submission/submission-service.ts
```diff
@@ -892,25 +896,59 @@ export namespace operations {
...
+                  };
+                  processedRecord = normalizedExceptionRecord;
+                } else {
+                  // only add validation errors that don't have exceptions
+                  validationErrors.push(validationError);
```
joneubank, line 937: I like the decision to do this but it is kind of subtle... You are normalizing the input to exactly match the normalized exception value. This deserves to be stated in a comment that we are doing this, probably near the top of this block or top of the method. Somethign should say that "Not only are we allowing the value through if there is an applicable exception, but we will normalize the value to exactly match the allowed value. Normalizing is setting it to start Upper case and to trim whitespace"... something like that

something

joneubank (commented):
I have no functional issues with this. I think it deserves more commenting for future readers, plus I left a couple comments for stylistic stuff.

src/submission/submission-service.ts
headsoup: you mean without '...{', right?

src/submission/submission-service.ts
ciaranschutte: went with default export, as you said no issue serverside plus its mostly used this way in the project

PR #946, opened 2023-04-04
feature branch: Entity exception

ciaranschutte:
**Description of changes**

Feature branch of changes that have been approved already

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/exception/error-handling.ts
```diff
@@ -0,0 +1,75 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { RepoError } from './repo/types';
+import { Errors } from '../utils';
+
+// types
+export type Success<T> = { success: true; exception: T };
+export type Failure = { success: boolean; message: string; errors?: any };
```
justincorrigible, line 25: just to clarify: for type `Success`, the property `success` is always true, but for type `Failure` it could be either true or false?

src/exception/exception-api.ts
```diff
@@ -17,87 +17,134 @@
...
-    }
+    const records = await parseTSV(req.file.path);
+    const programExceptionRecords = validateProgramExceptionRecords(records);
+
+    const result = await exceptionService.operations.createProgramException({
+      programId,
+      records: programExceptionRecords,
+    });
+
+    const status = !result.success ? 422 : 201;
```
justincorrigible, line 69: As personal policy, I wouldn’t not ask not to use unnecessary negative conditions 😂
```suggestion
    const status = result.success ? 201 : 422;
```

src/exception/error-handling.ts
ciaranschutte: nice catch! changed to a `boolean` from the constant `true`
types should be boolean , then the helper function sets as `true` or `false`

src/exception/exception-service.ts
```diff
@@ -17,104 +17,78 @@
...
 
-type Service = ({ programId }: { programId: string }) => Promise<Result>;
+/**
+ * normalize before schema validation
+ * tsv record values may contain different casing, validation needs normalized casing
+ * eg. tsv record value may be 'Follow Up', this will not pass schema validation for 'follow_up'
+ * @param records
+ */
+const normalizeRecords = (records: readonly EntityExceptionRecord[]) =>
+  records.map(r => ({
```
justincorrigible, line 67: my I suggest using explicit variable names? not in this function, but in more complicated ones, finding `r` may be difficult. ‘record’ would be both more legible and perfectly descriptive.

ciaranschutte: changing this from "r" to "record"
only place I could find the poor "r" naming, if there's another place you're looking at please let me know

src/exception/error-handling.ts
justincorrigible: That definitely works! Alternatively, one may have `true` and the other `false` as their single expected value, right?
I cannot imagine a successful Failure in this context (while I can in other more ridiculous cases 😆)

ciaranschutte: definitely works too for this. not relevant to this PR specifically. just an aside...

Avoiding the `true` explicit value because in the future there may be cases where one of these types takes something that has been cast into a boolean. I'm unclear if TS is ok with that or needs an explicit `true`. So this at least has that little bit of freedom.  I do see your point though.

justincorrigible: Agreed on all above! 🙏

src/exception/exception-service.ts
justincorrigible: I haven't found it anywhere else here, thus far. But having to refactor Arranger, where there's a lot of `x => ..`, I'm all for saving future maintainers preventable headaches.

src/exception/repo/program.ts
```diff
@@ -16,86 +16,69 @@
...
 const programExceptionSchema = new mongoose.Schema<ProgramException>({
   programId: String,
   exceptions: [
     {
+      program_name: String,
       schema: String,
-      coreField: String,
-      exceptionValue: { type: String, enum: Object.values(ExceptionValue) },
+      requested_core_field: String,
+      requested_exception_value: { type: String, enum: Object.values(ExceptionValue) },
```
justincorrigible, line 33: Not a request for change, but I'm curious about the change from camelCase.
Care to elaborate?

src/exception/repo/program.ts
```diff
@@ -16,86 +16,69 @@
...
     },
   ],
 });
 
-export const ProgramExceptionModel = mongoose.model<ProgramException>(
-  'ProgramException',
-  programExceptionSchema,
-);
+// check if model exists already to account for file watchers eg. test runner with live reload
```
justincorrigible, line 38: 🙏 thank you, this is helpful

src/exception/repo/program.ts
```diff
@@ -16,86 +16,69 @@
...
         { upsert: true, new: true, overwrite: true },
       ).lean(true);
-      // L.info(`doc created ${doc}`);
+      L.info(`doc created ${JSON.stringify(doc)}`);
+
+      return doc;
     } catch (e) {
       L.error('failed to create program exception: ', e);
-      return RepoError.SERVER_ERROR;
+      throw new DatabaseError('Cannot save program exception.');
```
justincorrigible, line 57: nice pattern... throwing simplified errors rather than system logs or stacks. 👍

src/exception/repo/program.ts
```diff
@@ -16,86 +16,69 @@
...
+  async find(programId: string): Promise<ProgramException | null> {
     L.debug(`finding program exception with id: ${JSON.stringify(programId)}`);
     try {
       const doc = await ProgramExceptionModel.findOne({ programId }).lean(true);
-      return checkDoc(doc);
+      return doc;
     } catch (e) {
       L.error('failed to find program exception', e);
-      return RepoError.SERVER_ERROR;
+      throw new DatabaseError('Cannot save program exception.');
```
justincorrigible, line 68: I think you've meant "cannot find"?
```suggestion
      throw new DatabaseError('Cannot find program exception.');
```

src/exception/repo/program.ts
```diff
@@ -16,86 +16,69 @@
...
+  async delete(programId: string): Promise<ProgramException | null> {
     L.debug(`deleting program exception with program id: ${JSON.stringify(programId)}`);
     try {
       const doc = await ProgramExceptionModel.findOneAndDelete({ programId }).lean(true);
-      return checkDoc(doc);
+      return doc;
     } catch (e) {
       L.error('failed to delete program exception', e);
-      return RepoError.SERVER_ERROR;
+      throw new DatabaseError('Cannot save program exception.');
```
justincorrigible, line 79: same here?
```suggestion
      throw new DatabaseError('Cannot delete program exception.');
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  async find(programId: string): Promise<EntityException | null> {
+    L.debug(`finding entity exception with program id: ${JSON.stringify(programId)}`);
+    try {
+      // first found document, or null
+      const doc = await EntityExceptionModel.findOne({ programId }).lean(true);
+      return doc;
+    } catch (e) {
+      L.error('failed to find program exception', e);
+      throw new DatabaseError();
```
justincorrigible, line 91: For consistency...
```suggestion
      throw new DatabaseError('Cannot find entity exception.');
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+        const doc = await entityExceptionDoc.save();
+        return doc;
+      }
+      // mongo will return nulls for non existent docs
+      // tslint doesn't complain about a null from a lib
+      // tslint:disable-next-line
+      return null;
+    } catch (e) {
+      L.error('failed to delete exception', e);
+      throw new DatabaseError('Cannot save entity exception.');
```
justincorrigible, line 127: Ok. maybe this was on purpose, but just in case...
```suggestion
      throw new DatabaseError('Cannot delete entity exception.');
```

src/exception/repo/types.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import { ObjectValues } from '../types';
+
+export type RepoResult<TException> = Promise<TException>;
+
+export const RepoError = {
+  DOCUMENT_UNDEFINED: 'DOCUMENT_UNDEFINED',
+  SERVER_ERROR: 'SERVER_ERROR',
+} as const;
```
justincorrigible, line 27: Looks like this could've been an `enum` instead.
However, I'm a bit iffy on which approach is best here.
Maybe @joneubank cares to chime in?

src/exception/types.ts
```diff
@@ -19,23 +19,59 @@
...
+  | ReadonlyArray<FollowUpExceptionRecord>;
+
+/**
+ * entity values to be valid EntityException exceptions arrays
+ * provide typing to tsv schema string to exception schema string
+ */
+export const EntityValues: Record<string, Exclude<keyof EntityException, 'programId'>> = {
+  specimen: 'specimen',
+  follow_up: 'follow_up',
+} as const;
```
justincorrigible, line 64: this is another case in which I'd have used an enum instead, but I remember @joneubank having strong opinions against it (I think?)

src/exception/repo/program.ts
ciaranschutte: TLDR: keeping as snake_case throughout reduced complexity and is consistent with code already present
two main things.
- our clinical entity records get saved as snakecase eg. `number_lymph_nodes_positive` so there is prior work that is snakecased
- tsv upload is snakecased, so there was a lot of additional code just to convert property names back and forth for the JS idiomatic camelCase.

src/exception/types.ts
```diff
@@ -65,7 +102,40 @@ export const isProgramExceptionRecord = (input: any): input is ProgramExceptionR
...
+export const isEntityExceptionRecord = (input: any): input is EntityExceptionRecord => {
+  return (
+    isExceptionRecordCheck(input) &&
+    (isSpecimenExceptionRecord(input) || isFollowupExceptionRecord(input))
+  );
+};
+
+const isExceptionRecord = (input: any): input is ExceptionRecord => isExceptionRecordCheck(input);
+
+export const isProgramExceptionRecord = isExceptionRecord;
```
justincorrigible, line 129: for backwards compatibility, I presume?
```suggestion
// TODO: rename this variable wherever else it's being used
export const isProgramExceptionRecord = isExceptionRecord;
```

ciaranschutte: to make things a little more defined.
`isExceptionRecord` is a base case for all exception records
it just so happens that `isProgramExceptionRecord` hasn't got any additional logic right now so they're one and the same.

in practice reading `isProgramExceptionRecord(...)` to check if something is a program exception record, is more clear than , `isExceptionRecord(...)`

tsconfig.json
```diff
@@ -22,5 +22,9 @@
   },
 
   "exclude": ["node_modules"],
-  "include": ["src/**/*", "test/**/*"]
+  "include": ["src/**/*", "test/**/*"],
+
+  "ts-node": {
+    "logError": true
```
justincorrigible, line 28: 👍

package.json
justincorrigible, line 1: fun fact, you can do `ts-node` in prod, using the `transpileOnly` flag.

src/exception/types.ts
justincorrigible: 🤔 clever. I see your point. Likely worth adding a comment about, so no hyperactive and overachieving future Justin will misunderstand the purpose and refactor it away!

src/exception/repo/types.ts
justincorrigible: caught up on that chat in slack...
https://www.youtube.com/watch?v=jjMbPt_H3RQ

src/exception/repo/program.ts
justincorrigible: You're right, we have to keep it consistently inconsistent.
Unfortunate, but not in the scope of this PR, as you well point out.

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    try {
+      const doc = await EntityExceptionModel.findOneAndUpdate(
+        { programId: exception.programId },
+        update,
+        { upsert: true, new: true, returnDocument: 'after' },
+      ).lean(true);
+      return doc;
+    } catch (e) {
+      L.error('failed to create entity exception: ', e);
```
justincorrigible, line 78:
```suggestion
      L.error('Failed to create entity exception: ', e);
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+      throw new DatabaseError('Cannot save entity exception.');
+    }
+  },
+
+  async find(programId: string): Promise<EntityException | null> {
+    L.debug(`finding entity exception with program id: ${JSON.stringify(programId)}`);
```
justincorrigible, line 84:
```suggestion
    L.debug(`Finding entity exception with program id: ${JSON.stringify(programId)}`);
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      // first found document, or null
+      const doc = await EntityExceptionModel.findOne({ programId }).lean(true);
+      return doc;
+    } catch (e) {
+      L.error('failed to find program exception', e);
```
justincorrigible, line 90:
```suggestion
      L.error('Failed to find program exception', e);
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+  },
+
+  async deleteSingleEntity(
+    programId: string,
+    entity: Entity,
+    submitterDonorIds: string[],
+  ): Promise<EntityException | null> {
+    L.debug(
+      `deleting single entity ${entity} exception with program id: ${JSON.stringify(programId)}`,
```
justincorrigible, line 101:
```suggestion
      `Deleting single entity ${entity} exception with program id: ${JSON.stringify(programId)}`,
```

src/exception/repo/entity.ts
```diff
@@ -0,0 +1,132 @@
...
+        entityExceptionDoc[entity] = filteredEntities as any;
+        const doc = await entityExceptionDoc.save();
+        return doc;
+      }
+      // mongo will return nulls for non existent docs
+      // tslint doesn't complain about a null from a lib
+      // tslint:disable-next-line
+      return null;
+    } catch (e) {
+      L.error('failed to delete exception', e);
```
justincorrigible, line 126:
```suggestion
      L.error('Failed to delete exception', e);
```

src/exception/util.ts
```diff
@@ -17,11 +17,42 @@
...
-import { ProgramException } from './types';
+import { RepoError } from './repo/types';
+import { EntityException, ProgramException, EntityValues, Entity } from './types';
+import _ from 'lodash';
 
-export function isProgramException(
-  result: ProgramException | RepoError,
-): result is ProgramException {
-  return (result as ProgramException).programId !== undefined;
+export function isProgramException(result: ProgramException | null): result is ProgramException {
```
justincorrigible, line 24: Question for my own TS understanding: why is `: result is ProgramException` necessary in this case?

src/submission/exceptions/exceptions.ts
```diff
@@ -0,0 +1,208 @@
...
+  programId: string;
+  record: DataRecord;
+  schemaName: ClinicalEntitySchemaNames;
+  schemaValidationErrors: dictionaryEntities.SchemaValidationError[];
+}) => {
+  const filteredErrors: dictionaryEntities.SchemaValidationError[] = [];
+  let normalizedRecord = record;
+
+  // retrieve submitted exceptions for program id (both program level and entity level)
+  const { programException, entityException } = await queryForExceptions(programId);
```
justincorrigible, line 55: Though this is not strictly speaking "wrong", I'm urged to suggest not using functions/variables before they're defined. I had to do `search` to find this in line 120, so I could find whether there's a try/catch associated with this await.

src/submission/exceptions/exceptions.ts
```diff
@@ -0,0 +1,208 @@
...
+  //   chemotherapy: 'chemotherapy',
+  //   immunotherapy: 'immunotherapy',
+  //   surgery: 'surgery',
+  //   radiation: 'radiation',
+  //   follow_up: 'followUps',
+  //   hormone_therapy: 'hormoneTherapy',
+  //   exposure: 'exposure',
+  //   comorbidity: 'comorbidity',
+  //   biomarker: 'biomarker',
+  //   sample_registration: 'sampleRegistration',
```
justincorrigible, line 204: future implementation?

justincorrigible (dismissed):
Preemptive approval 👍 
LGTM

src/exception/util.ts
ciaranschutte: it's the predicate type guard syntax https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/submission/exceptions/exceptions.ts
ciaranschutte: correct

src/exception/util.ts
justincorrigible: 🤔so it’s useful wherever the function returns, not within it.

got it! thanks 🙏

PR #948, opened 2023-04-04
Feat/860 get entity exception endpoint

ciaranschutte:
**Description of changes**

create api, service and route layers for retrieving entity exceptions endpoint

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/exception/exception-api.ts
+++ b/src/exception/exception-api.ts
@@ -118,6 +118,13 @@ class ExceptionController {
     return res.status(status).send(result);
   }
 
+  @HasFullReadAccess()
+  async getEntityException(req: Request, res: Response) {
+    const programId = req.params.programId;
+    const result = await exceptionService.operations.getEntityException({ programId });
+    return res.status(getResStatus(result)).send(result);
+  }
+
   @HasFullWriteAccess()
   async deleteEntityException(req: Request, res: Response) {
     const { programId } = req.params;
--- a/src/exception/exception-service.ts
+++ b/src/exception/exception-service.ts
@@ -172,6 +172,15 @@ export namespace operations {
     }
   };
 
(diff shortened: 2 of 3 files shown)


src/exception/exception-api.ts
```diff
@@ -118,6 +118,13 @@ class ExceptionController {
     return res.status(status).send(result);
   }
 
+  @HasFullReadAccess()
+  async getEntityException(req: Request, res: Response) {
+    const programId = req.params.programId;
```
joneubank, line 123: Any validation wanted on this param?

joneubank (approved):
LGTM

demariadaniel (approved):

src/exception/exception-service.ts
```diff
@@ -172,6 +172,15 @@ export namespace operations {
     }
   };
 
+  export const getEntityException: Service = async ({ programId }) => {
+    const result = await entityExceptionRepository.find(programId);
+
+    return processResult({
+      result,
+      errorMessage: `no entity level exceptions for program '${programId}'`,
```
demariadaniel, line 180: I'm a little thrown off with this errorMessage pattern, it looks like we're passing in our result and then this will be the default value if it's empty?

demariadaniel: b/c of isRepo Error I see ... I feel like this message should be returned from isRepoError given an empty result, not passed in as a default, but it's not crucial, just opinion

ciaranschutte: it's not a default though. the "repo error" is generic error for both program and entity exceptions.
`errorMessage` param is to have a more specific error message returned to user

PR #949, opened 2023-04-05
use tsv schema col for entity selection

ciaranschutte:
**Description of changes**

- submit other entity exception types
- take schema from submitted file tsv column

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/exception/exception-api.ts
+++ b/src/exception/exception-api.ts
@@ -18,18 +18,19 @@
  */
 
 import { NextFunction, Request, Response } from 'express';
+import _ from 'lodash';
 import { HasFullReadAccess, HasFullWriteAccess } from '../decorators';
 import { loggerFor } from '../logger';
 import { ControllerUtils, Errors, TsvUtils } from '../utils';
-import { RepoError } from './repo/types';
 import * as exceptionService from './exception-service';
+import { RepoError } from './repo/types';
 import {
-  isReadonlyArrayOf,
-  isProgramExceptionRecord,
-  isEntityExceptionRecord,
-  ProgramExceptionRecord,
   EntityExceptionRecord,
   EntityValues,
+  isEntityExceptionRecord,
+  isProgramExceptionRecord,
+  isReadonlyArrayOf,
+  ProgramExceptionRecord,
 } from './types';
 
 const L = loggerFor(__filename);
@@ -61,6 +62,13 @@ const validateEntityExceptionRecords: ValidateRecords<EntityExceptionRecord> = r
   }
   return records;
 };
+
+/**
+ * map submitted text from tsv schema column, to camelCase used in exception services
+ * @param record
+ * @returns camelCased schema name
+ */
+const getSchemaEntity = (record: EntityExceptionRecord) => EntityValues[_.camelCase(record.schema)];
 
 class ExceptionController {
   @HasFullWriteAccess()
@@ -108,10 +116,17 @@ class ExceptionController {
     const records = await parseTSV(req.file.path);
     const entityExceptionRecords = validateEntityExceptionRecords(records);
 
+    /**
+     * records have been validated
+     * schema is valid
+     * map tsv schema string to exception schema string
+     */
+    const schemaEntity = getSchemaEntity(entityExceptionRecords[0]);
+
     const result = await exceptionService.operations.createEntityException({
       programId,
       records: entityExceptionRecords,
-      entity: EntityValues.specimen,
+      entity: schemaEntity,
     });
 
     const status = !result.success ? 422 : 201;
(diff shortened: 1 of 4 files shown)


src/exception/repo/entity.ts
```diff
@@ -45,17 +56,17 @@ const EntityExceptionModel = mongoose.model<EntityException>(
...
   deleteSingleEntity(programId: string, entity: Entity): RepoResponse<EntityException>;
 }
 
 const entityExceptionRepository: EntityExceptionRepository = {
-  async save(exception: EntityException) {
+  async save(exception: OnlyRequired<EntityException, 'programId'>) {
     L.debug(`Creating new donor exception with: ${JSON.stringify(exception)}`);
 
-    const update = { $set: { programId: exception.programId, specimen: exception.specimen } };
+    const update = { $set: exception };
```
ciaranschutte, line 69: upserts, so if "followUp" is saved it will not overwrite "specimen"

src/exception/exception-api.ts
```diff
@@ -62,6 +63,13 @@ const validateEntityExceptionRecords: ValidateRecords<EntityExceptionRecord> = r
   return records;
 };
 
+/**
+ * map submitted text from tsv schema column, to camelCase used in exception services
+ * @param record
+ * @returns camelCased schema name
+ */
```
headsoup, line 70: 💪I dig this new stuff you're doing, dude

headsoup (dismissed):
didn't test it, but looking good!

src/exception/exception-api.ts
```diff
@@ -62,6 +63,13 @@ const validateEntityExceptionRecords: ValidateRecords<EntityExceptionRecord> = r
...
+const getSchemaEntity = (record: EntityExceptionRecord) => EntityValues[_.camelCase(record.schema)];
```
joneubank, line 71: So this is one of the places where TS let's us down. When you get the value out of `EntityValues` TS says that we are guaranteed to get one of the values from the const `EntityValues`... however `_.camelCase(record.schema)` could potentially be a value not available as a property of `EntityValues` (since `record.schema` is just a string`. So the actual return type has to be declares with `| undefined` otherwise we let errors leak in.

This is entirely a consequence of getting a value using `object[key]` type lookup. :(

src/exception/exception-api.ts
```diff
@@ -108,10 +116,17 @@ class ExceptionController {
     const records = await parseTSV(req.file.path);
     const entityExceptionRecords = validateEntityExceptionRecords(records);
 
+    /**
+     * records have been validated
+     * schema is valid
+     * map tsv schema string to exception schema string
+     */
+    const schemaEntity = getSchemaEntity(entityExceptionRecords[0]);
```
joneubank, line 124: Because of the previous comment, this could be `undefined`

joneubank (changes requested):
Going to continue reading but wanted to give you the feedback from my review of exception-api right away.

src/exception/types.ts
```diff
@@ -137,3 +141,6 @@ export const isReadonlyArrayOf = <T>(
 ): input is ReadonlyArray<T> => {
   return input.every(validator);
 };
+
+// utility
```
joneubank, line 145: This gets used frequently, it deserves a better explanation for any TS newby that interacts with it.

src/exception/exception-service.ts
```diff
@@ -29,29 +29,50 @@ import {
...
 const recordsToEntityException = ({
   programId,
   records,
 }: {
   programId: string;
   records: ReadonlyArray<EntityExceptionRecord>;
-  entity: Entity;
 }) => {
-  const exception: EntityException = { programId, specimen: [], followup: [] };
+  const exception: OnlyRequired<EntityException, 'programId'> = { programId };
```
joneubank, line 53: Good utility type usage.

src/exception/repo/entity.ts
```diff
@@ -45,17 +56,17 @@ const EntityExceptionModel = mongoose.model<EntityException>(
 );
 
 export interface EntityExceptionRepository {
-  save(exception: EntityException): RepoResponse<EntityException>;
+  save(exception: OnlyRequired<EntityException, 'programId'>): RepoResponse<EntityException>;
```
joneubank, line 59: I really like this type utility trick.

joneubank (commented):
The rest of this was good. If we can clean up that previous issue then this is good to go.

ciaranschutte:
@joneubank 
I've removed the entire `getSchemaEntity` function.
The `schema` value could be any `string` until after we validate the data here: https://github.com/icgc-argo/argo-clinical/blob/feat%2F859_submit_followup_exception/src/exception/exception-service.ts#L141
So it's not a trustworthy value.

The reason to get the schema early was to include it in an error message.
Changed the error message to just be generic base case
The error handling in general needs some cleanup before the exception feature gets merged to develop.
I would like to tackle that in a "error handling ticket"

joneubank (approved):

PR #955, opened 2023-04-14
Entity exception error handling

ciaranschutte:
**Description of changes**

- consistent and succinct error handling across Entity and Program exceptions
- merging into `entity_exception` feature branch
- reworks error handling for both Entity and Program exceptions 
  - success or failure types with suitable helper functions replace convoluted and duplicated processing 
  - thrown errors for validation, processing exceptions and errors
  - central Exception handling middleware registered on the route, idiomatic Express/node and consistent with already registered global middleware handling
  
 
**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/exception/error-handling.ts
+++ b/src/exception/error-handling.ts
@@ -21,12 +21,12 @@ import { RepoError } from './repo/types';
 import { Errors } from '../utils';
 
 // types
-export type Success<T> = { success: true; data: T };
+export type Success<T> = { success: true; exception: T };
 export type Failure = { success: boolean; message: string; errors?: any };
 export type Result<T> = Success<T> | Failure;
 
 // helpers
-export const success = <T>(data: T): Success<T> => ({ success: true, data });
+export const success = <T>(data: T): Success<T> => ({ success: true, exception: data });
 export const failure = (message: string, errors?: any): Failure => ({
   success: false,
   message,
--- a/src/exception/error-handling.ts
+++ b/src/exception/error-handling.ts
@@ -18,45 +18,58 @@
  */
 
 import { RepoError } from './repo/types';
+import { Errors } from '../utils';
 
 // types
 export type Success<T> = { success: true; data: T };
-// TODO: build out Failure to include error object eg. on POST
-export type Failure = { success: boolean; message: string };
+export type Failure = { success: boolean; message: string; errors?: any };
 export type Result<T> = Success<T> | Failure;
 
 // helpers
 export const success = <T>(data: T): Success<T> => ({ success: true, data });
-export const failure = (message: string) => ({ success: false, message });
+export const failure = (message: string, errors?: any): Failure => ({
+  success: false,
+  message,
+  errors,
+});
 
 // middleware
 export const ExceptionErrorHandler = (err: any, req: any, res: any, next: any) => {
-  console.log('Exception --- error handler --- midlleware');
-  console.log(JSON.stringify(err));
-  res.status(404).send('custom error handler ran');
-  // ----------
-  if (err instanceof DatabaseError) {
-    // err.statusCode + err.msg
+  const defaultErrorMessage = `Cannot create exceptions for entity in program '${req.params.programId}'`;
+
+  if (err instanceof ValidationError) {
+    const message = `${defaultErrorMessage}. Validation errors in exceptions file.`;
+    return res.status(404).send(failure(message, err.data));
   }
 
-  return res.status(err.status).send(err.data);
+  res.status(404).send(failure(err.message || defaultErrorMessage));
+
+  next(err);
 };
 
 // errors
 export class ValidationError extends Error {
+  data: ValidationError[];
   constructor(errors: any) {
-    super('message');
+    super();
     this.name = this.constructor.name;
-    // const errorMessage = `Cannot create exceptions for entity in program '${programId}'`;
+    this.data = errors;
   }
 }
 
 export class DatabaseError extends Error {
   constructor(message?: string, cause: RepoError = RepoError.SERVER_ERROR) {
-    super('Server error');
+    super(message || 'Failed to save exception.');
     this.name = this.constructor.name;
 
     if (cause === RepoError.DOCUMENT_UNDEFINED)
       this.message = 'Failed to save exception. Document is undefined';
+  }
+}
+
+export class ExceptionTSVError extends Errors.TSVParseError {
+  constructor(message?: string) {
+    super(message || 'TSV parsing error.');
+    this.name = this.constructor.name;
   }
 }
--- a/src/exception/error-handling.ts
+++ b/src/exception/error-handling.ts
@@ -21,7 +21,8 @@ import { RepoError } from './repo/types';
 
 // types
 export type Success<T> = { success: true; data: T };
-export type Failure = { success: false; message: string };
+// TODO: build out Failure to include error object eg. on POST
+export type Failure = { success: boolean; message: string };
 export type Result<T> = Success<T> | Failure;
 
 // helpers
--- a/src/exception/error-handling.ts
+++ b/src/exception/error-handling.ts
@@ -17,32 +17,39 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import * as express from 'express';
-import multer from 'multer';
-import exceptionApi, { requestContainsFile } from '../exception/exception-api';
-import { wrapAsync } from '../middleware';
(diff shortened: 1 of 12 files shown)


tsconfig.json
```diff
@@ -22,5 +22,10 @@
   },
 
   "exclude": ["node_modules"],
-  "include": ["src/**/*", "test/**/*"]
+  "include": ["src/**/*", "test/**/*"],
+
+  "ts-node": {
+    // server wont run if there are type errors, leave it to build step to typecheck
+    "transpileOnly": true
+  }
```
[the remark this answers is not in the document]
justincorrigible: if you use "logErrors" instead, it'll still show them without crashing, which is useful

https://github.com/icgc-argo/platform-api/blob/develop/tsconfig.json#L36

src/exception/exception-api.ts
```diff
@@ -84,24 +74,23 @@ class ExceptionController {
...
   @HasFullWriteAccess()
   async createEntityException(req: Request, res: Response) {
     const programId = req.params.programId;
-    const existingProgramException = await exceptionService.operations.getProgramException({
-      programId,
-    });
+    const doc = await programExceptionRepository.find(programId);
 
-    if (existingProgramException.exception !== undefined) {
+    if (doc && doc.exceptions !== undefined) {
```
justincorrigible, line 93: would this achieve the same?
```suggestion
    if (doc?.exceptions !== undefined) {
```

src/exception/exception-api.ts
```diff
@@ -115,15 +104,14 @@ class ExceptionController {
...
-    return res.status(status).send(result);
+    return res.status(200).send(result);
   }
 
   @HasFullReadAccess()
   async getEntityException(req: Request, res: Response) {
     const programId = req.params.programId;
     const result = await exceptionService.operations.getEntityException({ programId });
-    return res.status(getResStatus(result)).send(result);
+    return res.status(result.success ? 200 : 404).send(result);
```
justincorrigible, line 114: In all other responses, a `400` was the result of not having a `result.success`.
Is there a non-obvious purpose for this one being different?
Assuming there's none, I'd change this, both for consistency and semantics.

tsconfig.json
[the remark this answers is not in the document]
justincorrigible: on the other hand, that `transpileOnly` is a useful flag if we wanted to run ts-node in prod (like in `platform-api`)

src/exception/validation.ts
```diff
@@ -41,10 +41,12 @@ export const ValidationResultType = {
...
 type ValidationResultErrorType = ObjectValues<typeof ValidationResultType>;
 
-type ValidationError = {
+export type ValidationError = {
+  message: string;
+  result: ValidationResultErrorType;
   field: string;
   recordIndex: number;
-} & ValidationResult;
+};
```
justincorrigible, line 49: not a request for change: what's the benefit of spelling these properties out? it seems as if the `&` did the same job.

src/exception/repo/entity.ts
```diff
@@ -135,13 +116,15 @@ const entityExceptionRepository: EntityExceptionRepository = {
...
         const doc = await entityExceptionDoc.save();
-        return checkDoc(doc);
-      } else {
-        return RepoError.DOCUMENT_UNDEFINED;
+        return doc;
       }
+      // mongo will return nulls for non existent docs
+      // tslint doesn't complain about a null from a lib
+      // tslint:disable-next-line
+      return null;
```
justincorrigible, line 124: this return is preceded by an `if / else`, so it looks like it'd be unreachable and can be removed
```suggestion
```

src/exception/error-handling.ts
```diff
@@ -0,0 +1,75 @@
...
+  const defaultErrorMessage = `Cannot create exceptions for entity in program '${req.params.programId}'`;
+
+  if (err instanceof ValidationError) {
+    const message = `${defaultErrorMessage}. Validation errors in exceptions file.`;
+    return res.status(404).send(failure(message, err.data));
+  }
+
+  res.status(404).send(failure(err.message || defaultErrorMessage));
+
+  next(err);
```
justincorrigible, line 47: 👍 I like this pattern!
question: if you send a response in L45, what other function is this middleware passing the error forward to?
as the client will have closed the connection after this, is there an expectation to "send" another response?
or is it just to run some post-response process e.g. cleanup, or server-side logging?

tsconfig.json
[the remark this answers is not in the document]
ciaranschutte: nice!

src/exception/exception-api.ts
ciaranschutte: `    if (doc?.exceptions && doc.exceptions.length > 0) `
good call. yes the optional chaining works better. should be checking size of array too, not just if exists

src/exception/validation.ts
ciaranschutte: changed usage of types so they're separate distinct entities. 
they share those fields but ValidationError isn't linked to ValidationResult, in the sense that we don't want a change to ValidationResult changing the ValidationError type.

a small code smell in the structure of the validation overall is what is really surfacing here.

ValidationResult's come from individual validator functions
ValidationError's come from the overall validate that runs the individual validators

src/exception/exception-api.ts
justincorrigible: in that case `doc?.exceptions?.length > 0` would work too, right?

src/exception/repo/entity.ts
ciaranschutte: just preceded by an `if`
this is catching if no doc is found

src/exception/validation.ts
justincorrigible: in that case, both of those different types you mention could extend/inherit from a shared base type, no?

src/exception/error-handling.ts
ciaranschutte: no expectation. intent is just to pass it on to anything in the chain as best practice for anything that may come after, like your examples.

ciaranschutte: haven't tested it but for example I think you could attach an "error logger" middleware after this and log out just errors would be the way i see the logging example working

src/exception/exception-api.ts
ciaranschutte: `if (doc.exceptions && doc.exceptions.length > 0)` doing the same thing. no need for the optional chain on the doc?.exceptions

src/exception/exception-api.ts
justincorrigible: You're right. Personally, I agree that 400 is not the right approach here; as though it could work as a "give me something to work with, it does still sound more like an error than an empty response.

However, and based on the HTTP spec, I've always been under the impression 404 is descriptive of the server itself (https://www.rfc-editor.org/rfc/rfc9110.html#name-404-not-found); and it means "(server) resource not available", as in the endpoint itself is hidden or hasn't been implemented. A server "problem", not an "I could do my work with the endpoint you asked, but found no data". 404 gives the impression of a broken url (because that's what's semantically intended).

For that "no data", I believe the HTTP spec defined 204 "No content" (https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content).
from the spec: 
```
indicates that the server has successfully fulfilled the request and that there is no additional content to send in the response content
```
Not finding an exception could better be 204, as in 200 without anything to respond with.

src/exception/repo/entity.ts
justincorrigible: ah!! 😆 I saw the removed else and mistook it as part of the active code.
apologies!

src/exception/error-handling.ts
justincorrigible: yep. that's what I figured and double checking never hurts.
as I said, I like it!

ciaranschutte: appreciate this exact kind of engagement on PRs thank you

src/exception/exception-api.ts
justincorrigible: Found this as an interesting reference on the topic: https://apihandyman.io/empty-lists-http-status-code-200-vs-204-vs-404/. Being a big fan of semantics (to a pedantic fault), I would love to discuss it further.

src/exception/exception-api.ts
ciaranschutte: you're right but TS wants same types :)

ciaranschutte:
@justincorrigible updated to consistent 400 errors.
it's both consistent in this exception case and with the rest of the project, so should be ok for an approval.
please let me know if there's any other feedback and thanks for looking over this.

let's chat offline about best practices for the http status responses as I would like to hear more.
It would be helpful to put something up on the wiki about a "standard approach"

PR #980, opened 2023-05-16
Refactor Clinical API - Search Filters in request Body not Query

joneubank:
**Description of changes**

Clinical API search methods now take search filters in request Body instead of Query params. This was needed to accommodate large lists of donor IDs that are sometimes used in the submitted data tables.

It is unusual for a GET request to have a Body, but it is acceptable according to the HTTP spec and functions without issue.

Some refactoring to clean up usage of types was also done. Functionality should be unchanged other than the API formatting.

**Type of Change**

- [ ] Bug
- [x] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing


joneubank:
Still requires some testing and I need to update the Swagger docs. Tagging you now so you can take a look and see if I broke anything obvious...

src/clinical/api/clinical-api.ts
```diff
@@ -18,52 +18,33 @@
...
-  completionState?: {};
-};
-
-enum CompletionStates {
-  all = 'all',
-  invalid = 'invalid',
-  complete = 'complete',
-  incomplete = 'incomplete',
-}
-
```
demariadaniel, line 65: Ah some much needed improvements being made

src/resources/swagger.yaml
```diff
@@ -979,19 +982,26 @@ paths:
         '200':
           description: 'Array of Clinical Entity objects'
   /clinical/program/{programId}/clinical-errors:
-    get:
+    post:
       tags:
         - Clinical Data
       summary: Get all clinical errors for a given program. Returns clinicalErrors[].
+      description: This request fetches the migration errors attached to donors, filtered by the list of donor IDs provided in the request body. This does not update data any clinical data.
```
demariadaniel, line 989: Typo last sentence 'update data any clinical data'

src/submission/migration/migration-entities.ts
```diff
@@ -62,9 +62,5 @@ export type NewSchemaVerificationResult = {
 };
 
 export type DonorMigrationSchemaErrors = Array<
-  {
-    [clinicalAlias in EntityAlias | string]: ReadonlyArray<
-      dictionaryEntities.SchemaValidationError
-    >;
-  }
+  [ClinicalEntitySchemaNames, ReadonlyArray<dictionaryEntities.SchemaValidationError>]
```
demariadaniel, line 65: Changing from an object to an array here is giving me questions

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "skipLibCheck": true,
     "strict": true,
     "strictNullChecks": true,
-    "lib": ["esnext.array"],
+    "lib": ["ESNext"],
```
demariadaniel, line 10: I looked this up and seems like a good idea, we're going from a single specific esnext to the larger library, what motivated / forced this change?

src/submission/migration/migration-entities.ts
joneubank: The object being used was basically using an object with a single key/ value pair as a tuple, so i just made the whole thing into a tuple. This reduces complexity when using this data... for example we don't need to use `Object.entries` to access the key/value pair, and don't need to assume that only one key/value pair exists in the object... all that is baked into the tuple.

tsconfig.json
joneubank: I was looking for another issue in the tsconfig and thought this would be simpler to maintain. I don't have any concern using ESNext features and no specific reason to limit it to just the array lib. So basically it is simplifying.

src/submission/migration/migration-entities.ts
demariadaniel: That's great, had not considered that
Usages of DonorMigrationSchemaErrors will have to be updated down stream

src/clinical/api/clinical-api.ts
```diff
@@ -250,21 +229,30 @@ class ClinicalController {
...
     }
 
+    // 1. Get the migration errors for every donor requested...
+    const migrationErrors = await service.getClinicalEntityMigrationErrors(
+      programId,
+      bodyParseResult.data.donorIds,
+    );
+
+    // 2. ...and now remove from the list all valid donors (fixed with submissions since the migration)
+    const validDonorIds = await service.getDonorSubmissionErrorUpdates(programId, migrationErrors);
```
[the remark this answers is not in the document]
joneubank: I didn't rename this, it is taken from `req.params.programId` defined in the server. I agree we are inconsistent with this usage. If you want to do a quick refactor to make it consistent I'd be happy to approve that.

src/submission/migration/migration-entities.ts
joneubank: Are there other usages? looked like it was only used in one place. TS still compiles so I think we're safe. Do you know of places this could impact?

demariadaniel: Several unit tests fail as a result of this change, to me that indicates there could be additional impacts
https://jenkins.qa.cancercollaboratory.org/blue/organizations/jenkins/ARGO%2Fargo-clinical/detail/fix%2Fdata-query-large-filters/3/pipeline/

joneubank:
Pairs with changes in gateway here: https://github.com/icgc-argo/platform-api/pull/663

demariadaniel (approved):

Issue #854, opened 2022-11-28
Lost to Follow up field cross-file validation


hknahal:
This is related to https://github.com/icgc-argo/argo-dictionary/issues/358

A new optional field called `lost_to_followup_after_clinical_event_id` will be added to the Donor schema (see https://github.com/icgc-argo/argo-dictionary/blob/6de639fc8637e4843b203620fcda213b0cc8e03c/schemas/donor.json#L342) which will require cross-file validation checks in clinical.

### Clinical changes required:
If `lost_to_followup_after_clinical_event_id` is submitted:

1. The identifier submitted in the `lost_to_followup_after_clinical_event_id` must exist in the clinical submission in either the `submitter_primary_diagnosis_id` or `submitter_treatment_id` or `submitter_follow_up_id` field. This check will need to be done against the active submission and signed-off submission. If it does not match any of the existing IDs in the `submitter_primary_diagnosis_id` or `submitter_treatment_id` or `submitter_follow_up_id` fields, then the validation should fail and report an error message (eg. "The ID submitted in the 'lost_to_followup_after_clinical_event_id' does not exist in your clinical submission. Check your submission to ensure you submit the correct clinical event ID")

**Example of valid submission:**

The identifier submitted in the `lost_to_followup_after_clinical_even_id` field ("FL-22") exists in the `submitter_follow_up_id` column in the `Follow up` schema. This means the donor was last known to be alive 230 days after their primary diagnosis (this is the "interval_of_followup" value for submitter_follow_up_id "FL-22").
![ltf_valid](https://github.com/icgc-argo/argo-clinical/assets/4992044/95faa795-1f77-4303-947d-954c7aa3538b)

Test files:
NOTE: Change file suffix to *.tsv when using test files
[sample_registration.txt](https://github.com/icgc-argo/argo-clinical/files/11567373/sample_registration.txt)
[donor_valid.txt](https://github.com/icgc-argo/argo-clinical/files/11567081/donor_valid.txt)
[specimen_valid.txt](https://github.com/icgc-argo/argo-clinical/files/11567084/specimen_valid.txt)
[primary_diagnosis_valid.txt](https://github.com/icgc-argo/argo-clinical/files/11567083/primary_diagnosis_valid.txt)
[treatment_valid.txt](https://github.com/icgc-argo/argo-clinical/files/11567086/treatment_valid.txt)
[...]

PR #1002, opened 2023-05-29
Feat/854 Lost to Follow Up validation

demariadaniel:
**Description of changes**

[#854 ](https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/argo-clinical/854)
Validates submitted Donor records w/ `lost_to_followup` field:
a) treatment id referenced exists
b) new treatments are not submitted after `lost_to_followup` event

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing


test/integration/stub-schema.json
```diff
@@ -427,6 +427,16 @@
               "meta": {
                 "displayName": "Number of Children"
               }
+            },
+            {
+              "description": "If the donor became lost to follow up, indicate the identifier of the clinical event (eg. submitter_primary_diagnosis_id, submitter_treatment_id or submitter_follow_up_id) after which the donor became lost to follow up.",
+              "name": "lost_to_followup_after_clinical_event_id",
+              "valueType": "string",
+              "meta": {
```
demariadaniel, line 435: There should be `"restrictions": { "script": "#/script/donor/lost_to_followup" }` here as well but this line causes syntax errors for me

demariadaniel: Solved + updated

src/submission/submission-service.ts
```diff
@@ -571,13 +571,20 @@ export namespace operations {
     const filters: FindByProgramAndSubmitterFilter[] = [];
     // map records to submitterDonorId and build filters
     for (const clinicalType in existingActiveSubmission.clinicalEntities) {
-      const clinicalEnity = existingActiveSubmission.clinicalEntities[clinicalType];
-      clinicalEnity.records.forEach((rc, index) => {
+      const clinicalEntity = existingActiveSubmission.clinicalEntities[clinicalType];
```
justincorrigible, line 574: nice! 👍

src/submission/submission-to-clinical/merge-submission.ts
```diff
@@ -111,15 +111,15 @@ export const mergeActiveSubmissionWithDonors = async (
   return updatedDonors;
 };
 
-// This function will return a merged donor of records mapped by clinical type and the DB exsistentDonor
+// This function will return a merged donor of records mapped by clinical type and the DB existentDonor
 export const mergeRecordsMapIntoDonor = (
   unmutableSubmittedRecordsMap: DeepReadonly<SubmittedClinicalRecordsMap>,
-  exsistentDonor: DeepReadonly<Donor>,
+  existentDonor: DeepReadonly<Donor>,
```
justincorrigible, line 117: another nice catch 👍

src/submission/validation-clinical/donor.ts
```diff
@@ -42,11 +57,18 @@ export const validate = async (
...
   // other checks here and add to `errors`
+  const crossFileErrors = await crossFileValidator(
+    submittedDonorClinicalRecord,
+    submittedRecords,
+    mergedDonor,
+  );
+
+  errors = [...errors, ...crossFileErrors];
 
   return { errors };
```
justincorrigible, line 73: A bit confused about these changes here. The purpose of this logic seems unclear at first glance, so let me try to walk through it as I reason with it like a new team member would in a few months/years:

Initially, the empty `errors` array was a `const`, but needed to be changed into a `let`, presumably so it can be mutated in line 71. This can be avoided by simply returning a new array inline (see suggestion below).

However, now I wonder what the point of `errors` array itself was to begin with, defined as an empty array; of course, unless it's needed by `checkTimeConflictWithSpecimens` who may be mutating it elsewhere... which is both not obvious here, and possibly a source of rather hard to debug issues later on (and then again I'm speculating what is happening here, as I didn't go looking deeper into that function).
The code as it was doesn't really explain what it's meant to do.

Looks to me (guessing) like leaving the `const` unmodified, and just giving a different return for the whole function would be a cleaner change here; but that still doesn't clarify to me why the return needed an empty array to begin with, unless that other function is indeed mutating it elsewhere which is smelly code.

```suggestion
  const errors: SubmissionValidationError[] = []; // all errors for record
  // cross entity donor-specimen record validation
  checkTimeConflictWithSpecimens(submittedDonorClinicalRecord, mergedDonor, errors);

  // other checks here and add to `errors`
  const crossFileErrors = await crossFileValidator(
    submittedDonorClinicalRecord,
    submittedRecords,
    mergedDonor,
  );

  return { errors: [...errors, ...crossFileErrors] };
```

src/submission/validation-clinical/donor.ts
```diff
@@ -63,7 +85,7 @@ function checkTimeConflictWithSpecimens(
     return;
   }
   const specimenIdsWithTimeConflicts: string[] = [];
-  const donoSurvivalTime: number = Number(donorRecord[DonorFieldsEnum.survival_time]);
+  const donorSurvivalTime: number = Number(donorRecord[DonorFieldsEnum.survival_time]);
```
justincorrigible, line 88: wow, it's a spree. nice catch! 👍

src/submission/validation-clinical/donor.ts
demariadaniel: Yes, updating the return value is a great idea. 

So the back story here is a lot of the validation is using `.push` and it's across a bunch of functions, in this case `checkTimeConflictWithSpecimens`.

I really prefer keeping things functional/encapsulated so it's a mismatch of patterns btwn `crossFileErrors` and `checkTimeConflictWithSpecimens`

But now that I look at it, `checkTimeConflictWithSpecimens` is not exported or used anywhere else, so maybe there's a chance to refactor here.

src/common-model/functions.ts
```diff
@@ -127,7 +127,7 @@ export function getClinicalEntitiesFromDonorBySchemaName(
   const result = getClinicalObjectsFromDonor(donor, clinicalEntitySchemaName) as any[];
   const clinicalRecords = result
     .map((e: any) => {
-      if (e.clinicalInfo) {
+      if (e?.clinicalInfo) {
```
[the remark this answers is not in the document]
demariadaniel: Yeah `e` is entity, I'll update

src/submission/validation-clinical/donor.ts
justincorrigible: Oh, that's great to hear! I'll leave this in your hands then.
Thanks for hearing me out, Dan 🙏 

While Mutation is of course a valid and useful approach used by several companies and indy devs out there, as you may have found out in this app it tends to get messy over time. Therefor, I have a personal rule to avoid `let` altogether, and whenever I've found it being used, it is almost always a smell of brittle code.

Let me know when you've made whatever changes you see fit, and I'll gladly give this PR another look 💪

src/submission/validation-clinical/donor.ts
demariadaniel: I do what I can ;)

justincorrigible: much appreciated! every little bit counts 💯

src/submission/validation-clinical/donor.ts
demariadaniel: Totally agree, felt wrong adding a `let` ;)

src/common-model/functions.ts
```diff
@@ -126,9 +126,9 @@ export function getClinicalEntitiesFromDonorBySchemaName(
 ): ClinicalInfo[] {
   const result = getClinicalObjectsFromDonor(donor, clinicalEntitySchemaName) as any[];
   const clinicalRecords = result
-    .map((e: any) => {
-      if (e.clinicalInfo) {
-        return e.clinicalInfo as ClinicalInfo;
+    .map((entityRecord: any) => {
+      if (entityRecord?.clinicalInfo) {
+        return entityRecord.clinicalInfo as ClinicalInfo;
```
justincorrigible, line 131: 💪 awesome!

src/submission/validation-clinical/donor.ts
```diff
@@ -75,22 +96,171 @@ function checkTimeConflictWithSpecimens(
...
+    donorIds: [],
+    submitterDonorIds: [],
+  };
+
+  const { donors } = await donorDao.findByPaginatedProgramId(programId, query);
+
+  const storedTreatments = donors.map(donor => donor.treatments).filter(donor => Boolean(donor));
+  const storedPrimaryDiagnoses = donors
+    .map(donor => donor.primaryDiagnoses)
+    .filter(donor => Boolean(donor));
```
ciaranschutte, line 172: I'm not sure about the intent of this logic. 

The `filter` is getting an array of `primaryDiagnoses` not the `donor`.
Unclear on what the filter is doing as it's likely always true? `Boolean({})` and `Boolean(string[])` will be `true`.

src/submission/validation-clinical/donor.ts
```diff
@@ -75,22 +96,171 @@ function checkTimeConflictWithSpecimens(
...
+          : treatment;
+
+        const treatmentInterval =
+          treatment_start_interval && typeof treatment_start_interval === 'number'
+            ? treatment_start_interval
+            : 0;
+
+        const treatmentDuration =
+          treatment_duration && typeof treatment_duration === 'number' ? treatment_duration : 0;
+
```
ciaranschutte, line 236: these checks can be shortened because `typeof` will work on pretty much anything.
eg. `typeof this_is_not_a_valid_variable` will return 'undefined'

ciaranschutte:
couple of things, mostly in the `crossFileValidator` function.

1) `as` assertions are usually a sign that something isn't typed correctly or "flowing" through logic currently eg. a filter function not pulling out correct types. It would be great if we could fix these. I know they are everywhere in the codebase and platform-ui. I think as a team our TS knowledge has improved since then. 

2) There's a lot of filtering for `notEmpty` or `Boolean(<object>)`. Most of these arrays are actually constructed in this function so you have control over not having empty values in them from the start. This also helps not having to run filters on them multiple times. 

I think the filter logic can be cleaned up with a comment. On first look it appears like you were filtering with an object when actually the filter is just checking for empty strings. 
```javascript
const storedPrimaryDiagnoses = donors
    .map(donor => donor.primaryDiagnoses)
    .filter(donor => Boolean(donor));
 // to
 const storedPrimaryDiagnoses = donors
    .map(donor => donor.primaryDiagnoses)
    .filter(Boolean);
 // or
 const storedPrimaryDiagnoses = donors
    .map(donor => donor.primaryDiagnoses)
    .filter(primaryDiagnoses => Boolean(primaryDiagnoses));
```

3) Just trying to figure out the chat with you and Justin regarding preferring a functional pattern as I want to learn.
The code is still referring to the same object. It's just changing values in a different way.
using the `.push` accumulator style adds values to the array keeping the same object reference
using `let` and then just assigning a new array is keeping the same object reference too.

wouldn't functional pattern require a new object reference? Similar to react-redux pattern?
Where oldObject === newObject  is false?

ciaranschutte (changes requested):

src/common-model/functions.ts
```diff
@@ -126,9 +126,9 @@ export function getClinicalEntitiesFromDonorBySchemaName(
 ): ClinicalInfo[] {
   const result = getClinicalObjectsFromDonor(donor, clinicalEntitySchemaName) as any[];
   const clinicalRecords = result
-    .map((e: any) => {
-      if (e.clinicalInfo) {
-        return e.clinicalInfo as ClinicalInfo;
+    .map((entityRecord: any) => {
```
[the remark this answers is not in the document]
demariadaniel: Take a look at: [9f65de98](https://github.com/icgc-argo/argo-clinical/pull/1002/commits/9f65de9875c251d18c33ff95a4322203d2822892), this should do the trick

src/submission/submission-service.ts
```diff
@@ -571,13 +571,20 @@ export namespace operations {
...
+      clinicalEntity.records.forEach((rc, index) => {
```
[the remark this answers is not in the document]
demariadaniel: You're saying change `(rc, index)` to `(rc, record)` is that correct?

src/submission/validation-clinical/donor.ts
```diff
@@ -75,22 +96,171 @@ function checkTimeConflictWithSpecimens(
...
+const crossFileValidator = async (
+  submittedDonorRecord: DeepReadonly<SubmittedClinicalRecord>,
+  submittedRecords: DeepReadonly<ClinicalSubmissionRecordsByDonorIdMap>,
+  mergedDonor: DeepReadonly<Donor>,
+  // submitted records needed to validate current submission
+) => {
+  const { lost_to_followup_after_clinical_event_id, program_id } = submittedDonorRecord;
+  const { primaryDiagnoses, treatments, followUps } = mergedDonor;
+  const programShortName = program_id as string;
```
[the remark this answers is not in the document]
demariadaniel: This has been updated, the issue was coming from the Submitted Record types, which don't necessarily have the program as part of their value

Now the code pulls program from the current Donor, which always has program

src/submission/validation-clinical/donor.ts
demariadaniel: Updated

src/submission/submission-service.ts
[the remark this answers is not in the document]
demariadaniel: Actually not sure what you're saying here b/c the second argument in `forEach` is an `index`

We could update `rc` to use `record:
`clinicalEntity.records.forEach((record, index) => {
const donorId = record[SampleRegistrationFieldsEnum.submitter_donor_id];`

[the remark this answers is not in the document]
ciaranschutte: sorry my initial comment wasn't very descriptive.
suggesting changing `(rc, index)` to `(record, index)`

demariadaniel:
> couple of things, mostly in the `crossFileValidator` function.

@ciaranschutte Types + Filtering is updated + tested, please give a second look when you get a chance, thanks!

src/submission/submission-to-clinical/merge-submission.ts
```diff
@@ -111,15 +111,15 @@ export const mergeActiveSubmissionWithDonors = async (
...
 ) => {
   const submittedRecordsMap = _.cloneDeep(
     unmutableSubmittedRecordsMap,
   ) as SubmittedClinicalRecordsMap;
-  const mergedDonor: any = _.cloneDeep(exsistentDonor);
+  const mergedDonor: any = _.cloneDeep(existentDonor) || {};
```
ciaranschutte, line 122: is there a better fix for this than using `any`?
It looks like a workaround for `DeepReadOnly<obj>` vs `obj`

I know this issue is all over the codebase but I think it's a good opportunity to figure out a fix or at least a better bandage than `any` once and for all.

these `DeepReadOnly` types come from the `deep-freeze` package which is actually a straightforward singular function + typings. It just iterates over an object and recursively freezes it's properties using the js `freeze` function. Unfortunately in js this is the ultimate in locking things up ... there is no "unfreeze" (at least that I'm aware of).  Best way to work through this is to clone the object I think... which we are doing... but the frozen object types still follow. This is an issue because there is no enforcement of the freeze despite what the types are saying. You can see in this function the cloned object properties get mutated. 

I'll add a follow up comment as I'm currently trying to figure out the best approach here but wanted to get it started for any input you have

ciaranschutte: I think this is actually one of the few times using an assertion might be a valid case?
```
  const mergedDonor = (_.cloneDeep(existentDonor) || {}) as Donor;
```

we trust that `cloneDeep` is giving us a deep clone and it's a new object, and it's mutable. so we can tell TS that via the assertion.

ciaranschutte (approved):
couple of comments but looks for approval - thanks Dan

src/submission/submission-to-clinical/merge-submission.ts
demariadaniel: Comment #1: So the main purpose of this section is to compare `const treatmentInvalid = totalTreatmentTime > submittedInterval;` where `totalTreatmentTime` and `submittedInterval` should be numbers

`ClinicalInfo` is typed extremely generic, so there is no Type guarantee for variables such as `interval_of_followup`, `treatment_start_interval` and `treatment_duration`.

I actually think this section should be written as:

` const prevIntervalOfFollowUp =
        (typeof prevEntityIdMatch?.clinicalInfo?.interval_of_followup === 'number' &&
          prevEntityIdMatch.clinicalInfo.interval_of_followup) ||
        0;

      const submittedIntervalOfFollowUp =
        (typeof submittedIdMatch?.interval_of_followup === 'number' &&
          submittedIdMatch?.interval_of_followup) ||
        0;

      const submittedInterval = prevIntervalOfFollowUp
        ? prevIntervalOfFollowUp
        : submittedIntervalOfFollowUp;

      const prevInvalidTreatments = prevTreatments.filter(treatment => {
        const { treatment_start_interval, treatment_duration } = treatment.clinicalInfo
          ? treatment.clinicalInfo
          : treatment;

        const treatmentInterval =
          typeof treatment_start_interval === 'number' ? treatment_start_interval : 0;

        const treatmentDuration = typeof treatment_duration === 'number' ? treatment_duration : 0;

        const totalTreatmentTime = treatmentInterval + treatmentDuration;

        const treatmentInvalid = totalTreatmentTime > submittedInterval;`
        
        This guarantees the last line is comparing numberA > numberB

demariadaniel: see: https://github.com/icgc-argo/argo-clinical/pull/1002/commits/1d9e471ca8f65e3a68f2dbdd0b03276b679106be

ciaranschutte:
read your comments - sounds good - approving

one neat thing I came across was you can remove lines 137 to 139
then assign with a new variable name + default value at assignment time:

So you could remove this:
```typescript
const donorTreatments = typeof treatments !== 'undefined' ? treatments : [];
const donorPrimaryDiagnoses = typeof primaryDiagnoses !== 'undefined' ? primaryDiagnoses : [];
const donorFollowUps = typeof followUps !== 'undefined' ? followUps : [];
```
and add this
```typescript
  const {
    primaryDiagnoses: donorPrimaryDiagnoses = [],
    treatments: donorTreatments = [],
    followUps: donorFollowUps = [],
  } = mergedDonor;
```

no need to dismiss the code review to add this as I want to unblock you on this

ciaranschutte (approved):

demariadaniel:
> read your comments - sounds good - approving
> 
> one neat thing I came across was you can remove lines 137 to 139 then assign with a new variable name + default value at assignment time:
> 
> So you could remove this:
> 
> ```ts
> const donorTreatments = typeof treatments !== 'undefined' ? treatments : [];
> const donorPrimaryDiagnoses = typeof primaryDiagnoses !== 'undefined' ? primaryDiagnoses : [];
> const donorFollowUps = typeof followUps !== 'undefined' ? followUps : [];
> ```
> 
> and add this
> 
> ```ts
>   const {
>     primaryDiagnoses: donorPrimaryDiagnoses = [],
>     treatments: donorTreatments = [],
>     followUps: donorFollowUps = [],
>   } = mergedDonor;
> ```
> 
> no need to dismiss the code review to add this as I want to unblock you on this

Prefer this approach, thanks!

ciaranschutte (approved):

PR #1051, opened 2023-08-24
📊 1006 Clinical Data GQL

demariadaniel:
**Description of changes**

- Adds Resolvers for Clinical Data + Clinical Errors
  - nested ClinicalData / ClinicalErrors resolver structure is removed
- Updates to Clinical Errors:
  - Adds async handlers for Dictionary Validation
  - Makes ClinicalErrors non-nullable
  - Removes nested resolver + repeated async methods
- Adds Convert to GQL methods

Related changes:
Gateway: https://github.com/icgc-argo/platform-api/pull/672
UI: https://github.com/icgc-argo/platform-ui/pull/2713

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/clinical/clinical-service.ts
```diff
@@ -355,60 +386,76 @@ export const getValidRecordsPostSubmission = async (
...
-        errors: entityErrorRecords,
-      };
-
-      if (entityErrorRecords.length) clinicalErrors.push(errorResponseRecord);
-    });
+    const invalidDonorRecords = donorData.filter(
+      donor => !!donor.donorId && invalidDonorIds.includes(donor.donorId),
+    ) as DeepReadonly<DonorWithId>[];
+
+    const clinicalErrorRecords = await Promise.all(
```
demariadaniel, line 413: Revalidating records requires multiple dictionary validations

src/clinical/clinical-service.ts
```diff
@@ -44,23 +44,33 @@ import { runTaskInWorkerThread } from './service-worker-thread/runner';
 
 const L = loggerFor(__filename);
 
-export type ClinicalQuery = {
```
demariadaniel, line 47: Made Type names more specific to their purpose + return value

src/clinical/clinical-service.ts
```diff
@@ -44,23 +44,33 @@ import { runTaskInWorkerThread } from './service-worker-thread/runner';
...
 
-export type ClinicalSearchQuery = {
+// Returns paginated Donor + Entity Submission Data
+export interface PaginatedClinicalQuery extends ClinicalDonorEntityQuery {
+  page: number;
+  pageSize?: number;
+  sort: string;
+}
+
+// GQL Query Arguments
```
demariadaniel, line 63: Added new types for GQL request variables

src/dictionary/manager.ts
```diff
@@ -294,19 +294,23 @@ class SchemaManager {
...
 
-export const revalidateAllDonorClinicalEntitiesAgainstSchema = (
+export const revalidateAllDonorClinicalEntitiesAgainstSchema = async (
   donor: DeepReadonly<Donor>,
   schema: dictionaryEntities.SchemasDictionary,
 ) => {
   const clinicalSchemaNames = getSchemaNamesForDonorClinicalEntities(donor);
   let isValid = true;
-  clinicalSchemaNames.forEach((schemaName: ClinicalEntitySchemaNames) => {
+  for (const schemaName of clinicalSchemaNames) {
```
demariadaniel, line 303: Async only works properly in new `For of` loops

ciaranschutte: unclear where the async function is. Was this never working?

src/dictionary/manager.ts
```diff
@@ -294,19 +294,23 @@ class SchemaManager {
...
     }
-    const errs = MigrationManager.validateDonorEntityAgainstNewSchema(schemaName, schema, donor);
-    isValid = !errs || errs.length == 0;
-  });
+    const errs = await MigrationManager.validateDonorEntityAgainstNewSchema(
+      schemaName,
+      schema,
+      donor,
+    );
+    isValid = isValid && (!errs || errs.length == 0);
```
ciaranschutte, line 312: `isValid` is always going to be `true` here, so the original line is correct

src/clinical/clinical-service.ts
ciaranschutte: I think `interface extends type` is incorrect usage.
The way types extend is usually by unions.
I'm not sure why TS isn't throwing an error here.
Usually the main reason to use an interface is to extend from an interface.

What's the reasoning using the `interface` vs a `type`?
I ask because most of our code is written in types based on some talk we had a while back on slack with preferring types.

src/clinical/clinical-service.ts
```diff
@@ -313,6 +342,8 @@ export const getClinicalEntityMigrationErrors = async (
   return { migration, clinicalMigrationErrors };
 };
 
+type DonorWithId = Donor & { donorId: number };
```
ciaranschutte, line 345: `Donor` has a `donorId` property already
```

export interface Donor {
  _id?: string;
  __v?: number; // mongodb property not being filtered out
  createBy?: string;
  schemaMetadata: SchemaMetadata;
  donorId?: number;
  gender: string;
  submitterId: string;
  programId: string;
  specimens: Array<Specimen>;
  clinicalInfo?: ClinicalInfo;
  primaryDiagnoses?: Array<PrimaryDiagnosis>;
  familyHistory?: Array<FamilyHistory>;
  comorbidity?: Array<Comorbidity>;
  followUps?: Array<FollowUp>;
  treatments?: Array<Treatment>;
  exposure?: Array<Exposure>;
  biomarker?: Array<Biomarker>;
  createdAt?: string;
  updatedAt?: string;
  completionStats?: CompletionStats;
}
``` from src/clinical/clinical-entities.ts

File: src/clinical/clinical-service.ts [hunk 7]
```diff
@@ -355,60 +386,76 @@ export const getValidRecordsPostSubmission = async (
 
   const validDonorIds = donorData
     .filter(donor => donor.schemaMetadata.isValid)
-    .map(({ donorId }) => donorId);
-
-  const clinicalErrors: ClinicalErrorsResponseRecord[] = [];
+    .map(({ donorId }) => donorId)
+    .filter(id => typeof id === 'number');
```
ciaranschutte, line 390: Will `donorId` ever not be a number? 
If it's always a number no need to have an additional iteration with `filter`

src/clinical/clinical-service.ts
ciaranschutte, line 413: a comment would be helpful here. there is a lot to follow with multiple Promise arrays and nested mapping and then just returning the 1st index place of the array.

src/schemas/clinical-resolvers/clinicalData.ts
```diff
@@ -0,0 +1,80 @@
...
+  records: EntityDisplayRecord[][];
+}
+
+const convertClinicalDataToGql = (
+  programShortName: string,
+  clinicalEntities: ClinicalEntityData[],
+) => {
+  const clinicalDisplayData: ClinicalEntityDisplayData[] = clinicalEntities.map(
+    (entity: ClinicalEntityData) => {
+      const records: EntityDisplayRecord[][] = [];
```
ciaranschutte, line 29: what is the double `[][]` my google skills are failing me?

src/schemas/clinical-resolvers/clinicalData.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      entity.records.forEach((record: ClinicalInfo) => {
+        const displayRecords: EntityDisplayRecord[] = [];
+        for (const [key, val] of Object.entries(record)) {
+          const name = key === 'submitter_id' ? false : key;
```
ciaranschutte, line 34: having `name` as either a `string | false` seems like a type nightmare waiting to happen?

ciaranschutte (approved):
@demariadaniel left some questions & suggestions

src/clinical/clinical-service.ts
demariadaniel: But it's `donorId?` which is what's causing the issue

src/clinical/clinical-service.ts
demariadaniel: Apparently 🤷 Donor is typed with a `?`, so from Typescript's perspective no, donorId will not always be a number
I will dig into this a bit

src/clinical/clinical-service.ts
demariadaniel: Good call, will add

src/schemas/clinical-resolvers/clinicalData.ts
demariadaniel: An array containing arrays containing EntityDisplayRecords
`[ [ {} ], [ {} ], [ {} ] ]`

src/schemas/clinical-resolvers/clinicalData.ts
demariadaniel: Fair, will fix

src/dictionary/manager.ts
demariadaniel: Let me work back through this; dictionaryService.processRecords is the root of the issue and the fact that we're sometimes revalidating hundreds of records.

demariadaniel: But, may have got my wires crossed

src/clinical/clinical-service.ts
demariadaniel: Added

src/schemas/clinical-resolvers/clinicalData.ts
demariadaniel: OK updated, used a conditional `continue` to handle submitter_id
This original logic was written a long time ago, definitely worth an update

src/dictionary/manager.ts
demariadaniel: Right, cleaned up

src/clinical/clinical-service.ts
demariadaniel: Aha I see, still working on my Typescript. So where I got confused is `Type extends` is not a thing, `Type & {}` is the alternative. I used `interface` b/c I wanted to `extend` and didn't revisit.

src/clinical/clinical-service.ts
ciaranschutte: I see. the filter for checking number type threw me.
either it's an optional type and you check its not null in code. or it's a mandatory value.


we can do those checks rather than change the type. so you are right with a filter, it can just be a filter checking for existence of value. or setup default values, as I've seen in other code. 

just observations, totally your call on which way you want it to work

src/dictionary/manager.ts
demariadaniel: OK, I was able to de-Promisify across several files:
https://github.com/icgc-argo/argo-clinical/pull/1051/commits/5ee92de3cf23fd50c26925e17c9db86183550279

Debugging is a windy road. GQL was returning null values, but getClinicalErrors was returning correct values, so it appeared race condition was a possibility.

demariadaniel: Much improved this way

ciaranschutte (dismissed):

ciaranschutte (approved):
approving after merge. re ran tests fixed the failing build. something about a merge into head failing now? @demariadaniel  re-request if you need another approval otherwise I won't be notified

PR #1055, opened 2023-09-12
Clear clinical submission resolver

UmmulkiramR:
has the clearClinicalSubmission resolver
also has auth added to make user information available in the query.

### DIFF
--- a/src/schemas/clinical-resolvers/clearClinicalSubmissionResolver.ts
+++ b/src/schemas/clinical-resolvers/clearClinicalSubmissionResolver.ts
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2023 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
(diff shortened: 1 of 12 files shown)


src/server.ts
```diff
@@ -29,11 +31,16 @@ import { Server } from 'http';
...
 
-import { ApolloServer } from '@apollo/server';
-import { startStandaloneServer } from '@apollo/server/standalone';
+import { ApolloServer, ContextFunction } from '@apollo/server';
+import {
+  StandaloneServerContextFunctionArgument,
+  startStandaloneServer,
+} from '@apollo/server/standalone';
 import schema from './schemas/index';
+// import { EgoJwtData } from '@icgc-argo/ego-token-utils/dist/common';
```
demariadaniel, line 43: Can remove commented code?
I would include EgoJwtData with the GlobalGqlContext Type, which should have `userJwtData: EgoJwtData | null;`

src/schemas/clinical-resolvers/clearClinicalSubmissionResolver.ts
```diff
@@ -0,0 +1,43 @@
...
+    obj: unknown,
+    args: { programShortName: string; fileType: string; version: string },
+    contextValue: any,
+  ) => {
+    const { programShortName, fileType, version } = args;
+    const response = await submissionAPI.clearFileDataFromActiveSubmission(
+      programShortName,
+      fileType,
+      version,
+      (<GlobalGqlContext>contextValue).egoToken,
```
demariadaniel, line 35: For consistency w/ other functions I would change line 28 to `contextValue: GlobalGqlContext,` and 
line 35 to `contextValue.egoToken,`

UmmulkiramR: unfortunately, the _buildSubgraphSchema_ method does not accept resolvers of custom context type. Hence the typecast.

src/server.ts
UmmulkiramR: not sure why but argo clinical running locally is not able to fetch the type EgoJwtData from 
@icgc-argo/ego-token-utils/dist/common.

demariadaniel: `@icgc-argo/ego-token-utils` needs to be installed

src/schemas/clinical-resolvers/clearClinicalSubmissionResolver.ts
demariadaniel: OK I see now, sorry VSCode TS version was not set correctly

src/server.ts
UmmulkiramR: Oh yes! Done. Thanks.

demariadaniel (approved):

Issue #1031, opened 2023-08-04
1. Small Data Fetch : GQL Move - Platform to RDPC


bhavikbhagat93:
## Description
As part of federated argo, following GQL endpoints needs to be move to RDPC clinical service from platform gateway API

clinicalSubmissionTypesList
clinicalSubmissionSchemaVersion
clinicalSubmissionSystemDisabled

## Exit Criteria
Following resolvers should be available in the new RDPC clinical service once the implementation is completed.

PR #1060, opened 2023-09-20
#1031 Submission GQL Small Data Fetch

demariadaniel:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/dictionary/api.ts
+++ b/src/dictionary/api.ts
@@ -80,20 +80,22 @@ export const get = async (req: Request, res: Response) => {
   return res.status(200).send(schema);
 };
 
-export const getClinicalSchemas = async (withFields: boolean) =>
-  withFields
-    ? await manager
-        .instance()
+export const getClinicalSchemas = async (withFields: boolean) => {
+  const dictionaryManager = await manager.instance();
+  const schemaNameFilter = (schemaName: string) =>
+    schemaName !== ClinicalEntitySchemaNames.REGISTRATION;
+  const schemas = withFields
+    ? dictionaryManager
         .getSchemasWithFields()
-        .then(result =>
-          result.filter(
-            (s: manager.SchemaWithFields) => s.name !== ClinicalEntitySchemaNames.REGISTRATION,
-          ),
+        .then(schemaRecords =>
+          schemaRecords.filter(schemaRecord => schemaNameFilter(schemaRecord.name)),
         )
-    : await manager
-        .instance()
+    : dictionaryManager
         .getSchemaNames()
-        .then(result => result.filter(s => s !== ClinicalEntitySchemaNames.REGISTRATION));
+        .then(schemaRecords => schemaRecords.filter(schemaNameFilter));
+
+  return schemas;
+};
 
 export const getClinicalEntities = async (req: Request, res: Response) => {
   const withFields = req?.query?.includeFields?.toLowerCase() === 'true';
--- a/src/dictionary/api.ts
+++ b/src/dictionary/api.ts
@@ -80,16 +80,26 @@ export const get = async (req: Request, res: Response) => {
   return res.status(200).send(schema);
 };
 
+export const getClinicalSchemas = async (withFields: boolean) =>
+  withFields
+    ? await manager
+        .instance()
+        .getSchemasWithFields()
+        .then(result =>
+          result.filter(
+            (s: manager.SchemaWithFields) => s.name !== ClinicalEntitySchemaNames.REGISTRATION,
+          ),
+        )
+    : await manager
+        .instance()
+        .getSchemaNames()
+        .then(result => result.filter(s => s !== ClinicalEntitySchemaNames.REGISTRATION));
+
 export const getClinicalEntities = async (req: Request, res: Response) => {
-  const includeFields = req.query.includeFields as string;
-  if (includeFields && includeFields.toLowerCase() === 'true') {
-    const schemasWithFields = await manager.instance().getSchemasWithFields();
-    return res
-      .status(200)
-      .send(schemasWithFields.filter(s => s.name !== ClinicalEntitySchemaNames.REGISTRATION));
-  }
-  const schemas = await manager.instance().getSchemaNames();
-  return res.status(200).send(schemas.filter(s => s !== ClinicalEntitySchemaNames.REGISTRATION));
+  const withFields = req?.query?.includeFields?.toLowerCase() === 'true';
+  const schemas = await getClinicalSchemas(withFields).then(result => result);
+
+  return res.status(200).send(schemas);
 };
 
 export const getClinicalEntitiesData = async (includeFields: string) => {
--- a/src/dictionary/api.ts
+++ b/src/dictionary/api.ts
@@ -99,7 +99,7 @@ export const getClinicalSchemas = async (withFields: boolean) => {
 
 export const getClinicalEntities = async (req: Request, res: Response) => {
   const withFields = req?.query?.includeFields?.toLowerCase() === 'true';
-  const schemas = await getClinicalSchemas(withFields).then(result => result);
+  const schemas = await getClinicalSchemas(withFields);
 
   return res.status(200).send(schemas);
 };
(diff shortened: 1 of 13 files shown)


src/dictionary/api.ts
```diff
@@ -80,16 +80,26 @@ export const get = async (req: Request, res: Response) => {
   return res.status(200).send(schema);
 };
 
+export const getClinicalSchemas = async (withFields: boolean) =>
```
demariadaniel, line 83: This function has always returned either [String] or [{ }]; this does not play nicely w/ GraphQL TypeDefs
Open to suggestions + solutions

src/schemas/gqlTypeDefs.ts
```diff
@@ -339,6 +354,8 @@ const typeDefs = gql`
     oldValue: String!
     donorId: String!
   }
+
+  scalar SchemaList
```
demariadaniel, line 358: This scalar is a workaround for the issue in `/src/dictionary/api`
`clinicalSubmissionTypesList` cannot return `[String] | [SchemaWithFieldsObject]`; unions cannot include primitives

src/dictionary/api.ts
demariadaniel: Refactored for reusability (API vs GQL)
API is currently converting Clinical response to JSON, which masks difference btwn `[String] or [{ }]`

src/dictionary/api.ts
```diff
@@ -80,16 +80,26 @@ export const get = async (req: Request, res: Response) => {
...
+        .then(result =>
+          result.filter(
+            (s: manager.SchemaWithFields) => s.name !== ClinicalEntitySchemaNames.REGISTRATION,
+          ),
+        )
+    : await manager
+        .instance()
+        .getSchemaNames()
+        .then(result => result.filter(s => s !== ClinicalEntitySchemaNames.REGISTRATION));
+
```
ciaranschutte, line 97: is this logic the same apart from the instance method call?

could be little more DRY. not sure which is the async call so not sure where to put await..
``` javascript
export const getClinicalSchemas = async (withFields: boolean) => {
  const managerInstance = await manager.instance();
  const schemaNameFilter =  (schemaName) => schemaName !== ClinicalEntitySchemaNames.REGISTRATION;
  const schemas = withFields ? managerInstance.getSchemasWithFields().map(schema => schema.name) : managerInstance.getSchemaNames();
  return schemas.filter(schemaNameFilter)
}
```

src/dictionary/api.ts
```diff
@@ -80,16 +80,26 @@ export const get = async (req: Request, res: Response) => {
...
-  if (includeFields && includeFields.toLowerCase() === 'true') {
-    const schemasWithFields = await manager.instance().getSchemasWithFields();
-    return res
-      .status(200)
-      .send(schemasWithFields.filter(s => s.name !== ClinicalEntitySchemaNames.REGISTRATION));
-  }
-  const schemas = await manager.instance().getSchemaNames();
-  return res.status(200).send(schemas.filter(s => s !== ClinicalEntitySchemaNames.REGISTRATION));
+  const withFields = req?.query?.includeFields?.toLowerCase() === 'true';
+  const schemas = await getClinicalSchemas(withFields).then(result => result);
```
ciaranschutte, line 100: I think there's no need for `then`. 
there's two different Promise methods at work here.
the chaining `then` and the more modern `await` - they both (in a simple case) do the same thing
waiting until Promise resolves and doing something it.

src/schemas/clinical-mutations/clearRegistration.ts
```diff
@@ -22,26 +22,24 @@ import { loggerFor } from '../../logger';
...
+  const { shortName, registrationId } = args;
+  return await operations
+    .deleteRegistration(registrationId, shortName)
+    .then(() => {
+      L.info(`Deleted registrationId ${registrationId}`);
+      return true;
+    })
+    .catch(err => {
+      L.error(`Failed to delete registration`, err);
+      return false;
```
ciaranschutte, line 41: nice logging *thumbs up emoticon*

src/schemas/clinical-resolvers/clearClinicalSubmissionResolver.ts
```diff
@@ -21,23 +21,21 @@ import { GlobalGqlContext } from '../../app';
...
+  obj: unknown,
+  args: { programShortName: string; fileType: string; version: string },
+  contextValue: any,
+) => {
+  const { programShortName, fileType, version } = args;
+  const response = await submissionAPI.clearFileDataFromActiveSubmission(
+    programShortName,
+    fileType || 'all',
+    version,
+    (<GlobalGqlContext>contextValue).egoToken,
```
ciaranschutte, line 34: unclear why this `<GlobalGqlContext>` is needed? Casting just the contextValue?

src/schemas/clinical-resolvers/commitClinicalSubmission.ts
```diff
@@ -23,24 +23,22 @@ import { convertClinicalSubmissionDataToGql } from '../utils';
...
-      submission: submissionData,
-    });
-  },
+const commitClinicalSubmission = async (
+  obj: unknown,
+  args: { programShortName: string; version: string },
+  contextValue: any,
+) => {
+  const { programShortName, version } = args;
+  const submissionData = <DeepReadonly<ActiveClinicalSubmission>>(
```
ciaranschutte, line 32: is there one single way we can work around all this DeapReadonly stuff and stick with that approach?
I think this works out the same as using `as DeepReadonly....`
A utility function or a type guard that can be the de facto way for reusability, single point of functionality?

src/schemas/gqlTypeDefs.ts
ciaranschutte: Can you link me to more on this please? I don't understand and my gut is telling me that unions should be ok with primitives.

ciaranschutte (dismissed):
@demariadaniel left some questions & suggestions - approving

src/schemas/gqlTypeDefs.ts
demariadaniel: They aren't: https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/
Only allows Object types
Entire Github thread about it: https://github.com/graphql/graphql-spec/issues/215

demariadaniel: Going to sleep on it and review in the morning
In Gateway, this is simply `[String]` because the response array is converted to JSON, so that may be what client expects and I should just stick to the same format
I was trying to format this to see the proper object for GQL Server purposes, which I think is ideal but might require client side changes

ciaranschutte: sounds good - let me know where the union is happening tomorrow please - read the docs + github - very interesting - looks like unions don't work on scalars either 
```
All of a union's included types must be [object types](https://www.apollographql.com/docs/apollo-server/schema/schema/#object-types) (not scalars, input types, etc.). Included types do not need to share any fields.
```

src/dictionary/api.ts
demariadaniel: Yes I had a feeling this could be improved. But, we don't want to map `getSchemasWithFields()`, we need to return the objects with both names + fields, so `schemaNameFilter` needs to be called differently depending on value of withFields: https://github.com/icgc-argo/argo-clinical/pull/1060/commits/67ceec9e941a99262523a729233d3a15437078b5

src/dictionary/api.ts
demariadaniel: Yes this was unnecessary, removed

src/schemas/clinical-resolvers/commitClinicalSubmission.ts
demariadaniel: I think this one we can handle by casting the return type of `commitActiveSubmissionData`: `{} as DeepReadonly<ActiveClinicalSubmission>`

https://github.com/icgc-argo/argo-clinical/commit/7b08b5677bb7d588f3f43fe4dd133751909811f8

src/schemas/gqlTypeDefs.ts
demariadaniel: Oh so this Union is related to getSchemasWithFields: https://github.com/icgc-argo/argo-clinical/pull/1060#discussion_r1333305009
Where were either returning a String[] of Schema names or an array of SchemaWithFields objects

ciaranschutte (approved):

PR #1062, opened 2023-09-30
API Endpoint to Download TSV Files for donors by Donor ID

joneubank:
**Description of changes**

This adds a new API endpoint at `POST /clinical/donors/tsv` which accepts in the request body an array of donor IDs. It will collect all entity data for the requested donors and return that data in a Zip file with TSVs of each clinical entity.

This endpoint is restricted to Full Read Access only, requiring service level read access.

This will be used by the UI/API for users to request Clinical Data for files/donors that they have embargo level access to. Additional authorization restrictions for this request are managed by the Platform API Gateway.

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [x] Manual testing

### DIFF
--- a/src/clinical/api/clinical-api.ts
+++ b/src/clinical/api/clinical-api.ts
@@ -36,7 +36,10 @@ import {
   ClinicalErrorsApiBody,
   ClinicalSearchApiBody,
   CompletionState,
+  DonorDataApiBody,
 } from './types';
+import { loggerFor } from '../../logger';
+const L = loggerFor(__filename);
 
 // TODO: Update value type to match mongo schema
 export const completionFilters: Record<CompletionState, any> = {
@@ -136,6 +139,40 @@ class ClinicalController {
       .setHeader('content-disposition', fileName);
 
     const zip = createClinicalZipFile(entityData);
+
+    res.send(zip.toBuffer());
+  }
+
+  /**
+   * Download Clinical Data selected by Donor ID.
+   * Requires Full Read Access.
+   * Used by API for a user to download clinical data for specific donors and/or files.
+   * @param req
+   * @param res
+   * @returns
+   */
+  @HasFullReadAccess()
+  async getDonorDataByIdAsTsvsInZip(req: Request, res: Response) {
+    const bodyParseResult = DonorDataApiBody.safeParse(req.body);
+    if (!bodyParseResult.success) {
+      return ControllerUtils.badRequest(
+        res,
+        `Invalid data in request body: ${JSON.stringify(bodyParseResult.error)}`,
+      );
+    }
+    const { donorIds } = bodyParseResult.data;
+
+    const donorEntityData = await service.getDonorEntityData(donorIds);
+
+    const date = currentDateFormatted();
+    const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
+    res
+      .status(200)
+      .contentType('application/zip')
+      .attachment(fileName)
+      .setHeader('content-disposition', fileName);
+
+    const zip = createClinicalZipFile(donorEntityData);
 
     res.send(zip.toBuffer());
   }
--- a/src/clinical/clinical-service.ts
+++ b/src/clinical/clinical-service.ts
@@ -28,6 +28,7 @@ import {
   ClinicalErrorsResponseRecord,
   EntityAlias,
   aliasEntityNames,
+  allEntityNames,
 } from '../common-model/entities';
 import { Errors } from '../utils';
 import { patchCoreCompletionWithOverride } from '../submission/submission-to-clinical/stat-calculator';
@@ -53,12 +54,13 @@ export type ClinicalDonorEntityQuery = {
   completionState?: {};
 };
 
-// Returns paginated Donor + Entity Submission Data
-export type PaginatedClinicalQuery = ClinicalDonorEntityQuery & {
+export type PaginationQuery = {
   page: number;
   pageSize?: number;
   sort: string;
 };
+
+export type PaginatedClinicalQuery = ClinicalDonorEntityQuery & PaginationQuery;
 
 // GQL Query Arguments
 // Submitted Data Search Bar
@@ -120,6 +122,10 @@ export async function getDonors(programId: string) {
     return await donorDao.findByProgramId(programId);
   }
   return await donorDao.findBy({}, 999);
+}
+
+export async function getDonorsByIds(donorIds: number[]) {
+  return await donorDao.findByDonorIds(donorIds);
 }
 
 export async function findDonorId(submitterId: string, programId: string) {
@@ -219,14 +225,13 @@ export const getPaginatedClinicalData = async (
   query: PaginatedClinicalQuery,
 ) => {
   if (!programId) throw new Error('Missing programId!');
-  const start = new Date().getTime() / 1000;
 
   const allSchemasWithFields = await dictionaryManager.instance().getSchemasWithFields();
   // Get all donors + records for given entity
   const { donors, totalDonors } = await donorDao.findByPaginatedProgramId(programId, query);
 
   const taskToRun = WorkerTasks.ExtractEntityDataFromDonors;
-  const taskArgs = [donors as Donor[], totalDonors, allSchemasWithFields, query];
+  const taskArgs = [donors as Donor[], totalDonors, allSchemasWithFields, query.entityTypes, query];
 
   // Return paginated data
   const data = await runTaskInWorkerThread<{ clinicalEntities: ClinicalEntityData[] }>(
@@ -234,10 +239,30 @@ export const getPaginatedClinicalData = async (
     taskArgs,
   );
 
-  const end = new Date().getTime() / 1000;
-  L.debug(`getPaginatedClinicalData took ${end - start}s`);
-
   return data;
+};
+
+export const getDonorEntityData = async (donorIds: number[]) => {
+  const allSchemasWithFields = await dictionaryManager.instance().getSchemasWithFields();
+  // Get all donors + records for given entity
+  const donors = await getDonorsByIds(donorIds);
+  const totalDonors = donors.length;
+
+  const paginationQuery: PaginationQuery = {
+    page: 0,
+    sort: 'donorId',
+    pageSize: totalDonors,
+  };
+
+  const taskToRun = WorkerTasks.ExtractEntityDataFromDonors;
+  const taskArgs = [donors, donors.length, allSchemasWithFields, allEntityNames, paginationQuery];
+
+  // Return paginated data
+  const data = await runTaskInWorkerThread<{ clinicalEntities: ClinicalEntityData[] }>(
+    taskToRun,
+    taskArgs,
+  );
+  return data.clinicalEntities;
 };
 
 export const getClinicalSearchResults = async (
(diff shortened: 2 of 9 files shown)


src/clinical/clinical-service.ts
```diff
@@ -53,13 +54,14 @@ export type ClinicalDonorEntityQuery = {
...
 
-// Returns paginated Donor + Entity Submission Data
-export type PaginatedClinicalQuery = ClinicalDonorEntityQuery & {
+export type PaginationQuery = {
   page: number;
   pageSize?: number;
   sort: string;
 };
 
+export type PaginatedClinicalQuery = ClinicalDonorEntityQuery & PaginationQuery;
```
joneubank, line 63: We had a need to separate the a worker thread task to accept Pagination parameters separate from the Clinical Donor Entity query data, so we can separate out that data for reference to its own type `PaginationQuery`.

src/clinical/clinical-service.ts
```diff
@@ -219,27 +225,46 @@ export const getPaginatedClinicalData = async (
...
   if (!programId) throw new Error('Missing programId!');
-  const start = new Date().getTime() / 1000;
 
   const allSchemasWithFields = await dictionaryManager.instance().getSchemasWithFields();
   // Get all donors + records for given entity
   const { donors, totalDonors } = await donorDao.findByPaginatedProgramId(programId, query);
 
   const taskToRun = WorkerTasks.ExtractEntityDataFromDonors;
-  const taskArgs = [donors as Donor[], totalDonors, allSchemasWithFields, query];
+  const taskArgs = [donors as Donor[], totalDonors, allSchemasWithFields, query.entityTypes, query];
```
joneubank, line 234: the `ExtractEntityDataFromDonors` task was updated torequire the `PaginatedQuery` in place of the `ClinicalDonorEntityQuery` since the new request by DonorId doesn't include much of that request data. The only additional information used from that query type was the list of entities to include in the request, which is now passed as a separate argument `query.entityTypes`.

src/clinical/api/clinical-api.ts
```diff
@@ -140,6 +143,40 @@ class ClinicalController {
...
+        res,
+        `Invalid data in request body: ${JSON.stringify(bodyParseResult.error)}`,
+      );
+    }
+    const { donorIds } = bodyParseResult.data;
+
+    const donorEntityData = await service.getDonorEntityData(donorIds);
+
+    const date = currentDateFormatted();
+    const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
```
joneubank, line 168: This is sharing the naming convention used for for downloads requested by program submitters. I'm open to suggestions to improve the provided filename. Perhaps a timestamp is a good idea in addition to the current date.

src/clinical/api/clinical-api.ts
```diff
@@ -36,7 +36,10 @@ import {
   ClinicalErrorsApiBody,
   ClinicalSearchApiBody,
   CompletionState,
+  DonorDataApiBody,
 } from './types';
+import { loggerFor } from '../../logger';
+const L = loggerFor(__filename);
```
justincorrigible, line 42: Could `Log`, or `Logger` do the same job, while being a bit more declarative than a single capital letter `l`?

src/clinical/api/clinical-api.ts
```diff
@@ -140,6 +143,40 @@ class ClinicalController {
...
+    const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
+    res
+      .status(200)
+      .contentType('application/zip')
+      .attachment(fileName)
+      .setHeader('content-disposition', fileName);
+
+    const zip = createClinicalZipFile(donorEntityData);
+
+    res.send(zip.toBuffer());
```
justincorrigible, line 177: funny, I'd have imagined you'd get an error/warning here saying that you had already sent a response in line 170.
obviously I must be mistaken.

src/clinical/api/clinical-api.ts
joneubank: yes and it would be better. I'm being consistent with the existing pattern in this repo. I would - recommend a separate action to rename them all.

src/clinical/api/clinical-api.ts
joneubank: previous use of `res` didn't send data, just set some headers and status.

src/clinical/clinical-service.ts
justincorrigible: I see the type for query implies the property (`entityTypes`) is guaranteed to be present.
is it safe to assume its actual presence is validated on runtime elsewhere?

src/clinical/api/clinical-api.ts
joneubank: > aside from the filename format itself, would it make sense to abstract this into an env variable?

I don't see a reason to have different filenames per env, what's your reason? If you want filename constanta centralized into a constants file I can do that.

justincorrigible: Tbh, I'm just used to abstracting constants like this one, because that makes it easier for these values to be updated, without having to rebuild the code "just for a string change". 

If you believe this doesn't need to be abstracted, I of course trust your insight.

joneubank: Good reason. will make the change.

src/clinical/api/clinical-api.ts
justincorrigible: follow up: https://expressjs.com/en/api.html#res.status
looks like I've been away from coding for too long 😢 
...I miss creating things

src/clinical/clinical-service.ts
```diff
@@ -122,6 +124,10 @@ export async function getDonors(programId: string) {
   return await donorDao.findBy({}, 999);
 }
 
+export async function getDonorsByIds(donorIds: number[]) {
+  return await donorDao.findByDonorIds(donorIds);
```
justincorrigible, line 128: is this `await` necessary here? the function being async means the return will be a promise regardless, and the result can (or should?) be awaited there.

I recall seeing this be pointed out by ESLint at some point, but maybe I'm confusing things.

joneubank: nice catch. not sure if makes a difference but yes this await isn't needed

src/clinical/clinical-service.ts
```diff
@@ -219,27 +225,46 @@ export const getPaginatedClinicalData = async (
...
   );
 
-  const end = new Date().getTime() / 1000;
-  L.debug(`getPaginatedClinicalData took ${end - start}s`);
-
   return data;
 };
 
+export const getDonorEntityData = async (donorIds: number[]) => {
+  const allSchemasWithFields = await dictionaryManager.instance().getSchemasWithFields();
```
justincorrigible, line 246: no need to do anything about this one, just felt the need to acknowledge out loud how nervous it makes me to have these long chains without `?.` 😆

src/clinical/clinical-service.ts
```diff
@@ -219,27 +225,46 @@ export const getPaginatedClinicalData = async (
...
+
+  const taskToRun = WorkerTasks.ExtractEntityDataFromDonors;
+  const taskArgs = [donors, donors.length, allSchemasWithFields, allEntityNames, paginationQuery];
+
+  // Return paginated data
+  const data = await runTaskInWorkerThread<{ clinicalEntities: ClinicalEntityData[] }>(
+    taskToRun,
+    taskArgs,
+  );
+  return data.clinicalEntities;
```
justincorrigible, line 265: Out of habit, I always separate the functions' `return`, so it's easier to discern it from the rest of the logic.
```suggestion

  return data.clinicalEntities;
```

src/clinical/donor-repo.ts
```diff
@@ -392,7 +393,17 @@ export const donorDao: DonorRepository = {
       submitterId: { $in: submitterIds },
       programId: programId,
     });
-    const mapped = result.map((d: DonorDocument) => {
+    const mapped = result.map(d => {
+      return MongooseUtils.toPojo(d) as Donor;
```
justincorrigible, line 397: you, using `as`? le wut!

src/clinical/donor-repo.ts
```diff
@@ -392,7 +393,17 @@ export const donorDao: DonorRepository = {
...
+    });
+    return F(mapped);
```
justincorrigible, line 399: what on earth is `F`!?

...I know it's just the established pattern, and it's obviously a function, but what does it do!?
without having to change the whole app, at least the import could be renamed in this file.

src/clinical/donor-repo.ts
joneubank: Old mongoose code, there are better TS ways to do this now. Following convention of the file instead of investing time in a re-write of the DB management.

src/clinical/donor-repo.ts
joneubank: rewrites the type as `DeepReadonly<T>`. Again, following the service convention, I do not recommend adopting this pattern elsewhere.

justincorrigible: `import { F as Freeze }` should help... but even then, what's `freezing`? 😮‍💨

justincorrigible (approved):
LGTM (except for the deeper "pattern" issues pointed out in the comments, which are beyond the scope of this PR)

src/clinical/donor-repo.ts
ciaranschutte: can't update/extend properties dynamically
if I remember correctly it's a very light lib around the native Object.freeze() https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

src/clinical/clinical-service.ts
ciaranschutte: long optional chaining makes me nervous too.
I actually really like how `dictionaryManager` does this - throws an error if there's no instance - so you could actually just catch the error and return ... empty array of donorIds in that case?

that node pattern of throw/catch actual errors vs just retuning nulls/undefined is used a lot in this repo and I think it works well.

ciaranschutte: re `?` part of this.... `getSchemaWIthFields` will never actually be run on undefined, an error will be thrown instead

src/clinical/donor-repo.ts
ciaranschutte: screenshotted to use without context in the future (context provider is undefineed) :p

src/clinical/clinical-service.ts
justincorrigible: welp! sounds like that requires a lot of "insider" knowledge on that `dictionaryManager`.

try+catch sure does sound desirable, but I guess my non-evident point is that this line is a bit "code golf"ey, and could use being broken into separate lines 🤷‍♂️

joneubank: don't need insider knowledge... the type system informs is of what the output will be. if we can trust the type system then we're safe here.

justincorrigible: oh fun! didn't realise TS was visible in PR reviews at Github. must get myself that browser extension!

joneubank: sarcasm acknowledged. agreed that the lack of introspection in the browser code review is limiting.

justincorrigible: to be fair, we could review PRs by testing them locally... like we used to back in the day.

Modern tooling has become a crutch for what could be more declarative; which proves hard to keep up in the review process. How do you suggest we improve that?

PR #1065, opened 2023-10-10
Clinical submission upload api

UmmulkiramR:
The submission upload will be accessible via a rest API instead of GraphQL.
There is also some code refactoring added


src/submission/submission-api.ts
```diff
@@ -386,4 +433,18 @@ const isValidCreateBody = (req: Request, res: Response): boolean => {
   return true;
 };
 
+const isValidRequestArgs = (programId: string, uploadedFiles: {}): boolean => {
+  if (programId === undefined) {
+    L.debug('programId missing');
+    // ControllerUtils.badRequest(res, `programId is required`);
+    return false;
+  }
+  if (/*uploadedFiles === undefined &&*/ Object.keys(uploadedFiles).length === 0) {
```
demariadaniel, line 442: Looks like some test code that wasn't removed?

UmmulkiramR: Cleared it. Thanks.

src/schemas/clinical-resolvers/uploadClinicalSubmissions.ts
```diff
@@ -0,0 +1,57 @@
...
+    status = 207;
+  }
+
+  const response = await convertClinicalSubmissionDataToGql(programShortName, {
+    submission: uploadResult?.submission,
+    batchErrors: uploadResult?.batchErrors,
+    successful: uploadResult?.successful,
+  });
+
+  return res.status(status).send(response);
```
demariadaniel, line 54: Code for returning the response should be limited to the API; the resolver should only return the result of `await convertClinicalSubmissionDataToGql`

UmmulkiramR: The _uploadClinicalSubmissions_ adds a layer between the request and the code that processes the request. There is stuff to do to unpack the request and pack the response for the caller which can be handled here. I think it is a good idea to keep it this way.

UmmulkiramR:
cancelling as this was covered in [#1085](https://github.com/icgc-argo/argo-clinical/pull/1085)

PR #1108, opened 2024-01-19
Feat/1083 Validate Blank Numeric Exception Fields

demariadaniel:
**Description of changes**

Checks Exception Field value type (numeric vs string)
Validates submitted record against Program + Entity exceptions for numeric fields

Test files and validation results:
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/1083

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -65,7 +65,7 @@ const validateFieldValueWithExceptions = ({
   fieldValue: string;
   validationErrorFieldName: string;
 }): boolean => {
-  const allowedValues: Set<string> = new Set();
+  const allowedValues: Set<string | undefined> = new Set();
 
   // program level is applicable to ALL donors
   if (programException) {
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -206,16 +206,6 @@ export const checkForProgramAndEntityExceptions = async ({
       (valueType === 'number' || valueType === 'integer') &&
       isValidNumericExceptionType(fieldValue);
 
-    // missing required field, validate as normal, exceptions still require a submitted value
-    const isMissingRequiredField =
-      validationError.errorType ===
-      dictionaryEntities.SchemaValidationErrorTypes.MISSING_REQUIRED_FIELD;
-
-    if (isMissingRequiredField) {
-      filteredErrors.push(validationError);
-      return;
-    }
-
     let normalizedFieldValue = '';
     let normalizedValue: string | string[] = '';
 
@@ -228,7 +218,7 @@ export const checkForProgramAndEntityExceptions = async ({
 
       normalizedFieldValue = normalizedString;
     } else if (!validNumericExceptionValue) {
-      // Field value is a type we cannot allow exceptions for, validate as normal
+      // If field value is not string or number, then value is not a type we allow exceptions for
       filteredErrors.push(validationError);
       return;
     }
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -64,11 +64,11 @@ const validateFieldValueWithExceptions = ({
   programException: ProgramException | null;
   entityException: EntityException | null;
   schemaName: ClinicalEntitySchemaNames;
-  fieldValue: string;
+  fieldValue: string | undefined;
   validationErrorFieldName: string;
   valueType?: dictionaryEntities.ValueType;
 }): boolean => {
-  const allowedValues: Set<string> = new Set();
+  const allowedValues: Set<string | undefined> = new Set();
 
   const isNumericField =
     valueType === dictionaryEntities.ValueType.INTEGER ||
@@ -84,7 +84,7 @@ const validateFieldValueWithExceptions = ({
       )
       .forEach(matchingException => {
         if (isNumericField) {
-          allowedValues.add('');
+          allowedValues.add(undefined);
         } else {
           allowedValues.add(matchingException.requested_exception_value);
         }
@@ -116,7 +116,7 @@ const validateFieldValueWithExceptions = ({
       .filter(exception => exception.requested_core_field === validationErrorFieldName)
       .forEach(matchingException => {
         if (isNumericField) {
-          allowedValues.add('');
+          allowedValues.add(undefined);
         } else {
           allowedValues.add(matchingException.requested_exception_value);
         }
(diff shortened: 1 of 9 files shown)


src/submission/exceptions/exceptions.ts
```diff
@@ -161,52 +189,54 @@ export const checkForProgramAndEntityExceptions = async ({
...
   // if there are submitted exceptions for this program, check if they match record values
   if (!(programException || entityException)) {
-    return { filteredErrors: schemaValidationErrors, normalizedRecord: record };
+    return { filteredErrors: schemaValidationErrors, normalizedRecord };
   }
 
   // check each validation error for a matching exception, and remove the validaiton if the value matches the exception value
   schemaValidationErrors.forEach(validationError => {
     const validationErrorFieldName = validationError.fieldName;
     const fieldValue = record[validationErrorFieldName];
```
demariadaniel, line 198: For numeric values with blank fields, `fieldValue` is `undefined` at this point. Will update so integer values are not 'normalized', and `undefined` is stored in the DB

src/submission/exceptions/exceptions.ts
```diff
@@ -75,7 +82,13 @@ const validateFieldValueWithExceptions = ({
           exception.requested_core_field === validationErrorFieldName &&
           exception.schema === schemaName,
       )
-      .forEach(matchingException => allowedValues.add(matchingException.requested_exception_value));
+      .forEach(matchingException => {
+        if (isNumericField) {
+          allowedValues.add(undefined);
```
demariadaniel, line 87: Not sure if '' (empty string) should be included here as well

src/submission/exceptions/exceptions.ts
```diff
@@ -57,15 +58,21 @@ const validateFieldValueWithExceptions = ({
...
+  fieldValue: string | undefined;
   validationErrorFieldName: string;
+  valueType?: dictionaryEntities.ValueType;
 }): boolean => {
-  const allowedValues: Set<string> = new Set();
+  const allowedValues: Set<string | undefined> = new Set();
+
+  const isNumericField =
+    valueType === dictionaryEntities.ValueType.INTEGER ||
+    valueType === dictionaryEntities.ValueType.NUMBER;
```
joneubank, line 75: this would be a nice reusable function:

```ts
const isNumericField = (valueType: string): boolean => 
  valueType === dictionaryEntities.ValueType.INTEGER ||
  valueType === dictionaryEntities.ValueType.NUMBER;
```

src/submission/exceptions/exceptions.ts
joneubank: `undefined` is the correct value here. If we allow `''` we also need to convert it into `undefined` eventually. The way this code is strucutred, i think the value normalization happens before this validation code so we would want that conversion from empty string to `undefined` to happen outside of this function.

src/submission/exceptions/exceptions.ts
```diff
@@ -161,52 +189,56 @@ export const checkForProgramAndEntityExceptions = async ({
...
+    const validNumericExceptionValue =
+      (valueType === 'number' || valueType === 'integer') &&
+      isValidNumericExceptionType(fieldValue);
 
-    // missing required field, validate as normal, exceptions still require a submitted value
-    const isMissingRequiredField =
-      validationError.errorType ===
-      dictionaryEntities.SchemaValidationErrorTypes.MISSING_REQUIRED_FIELD;
+    let normalizedFieldValue: string | undefined = '';
+    let normalizedValue: string | string[] = '';
```
joneubank, line 208: Looks like we are checking the numeric field type correctly and then just setting the normalized value to `''` anyways.

This and the following section of code seem to be doing two things: 
1. validating the current fieldValue is of the correct type (`''` or `undefined` for numeric, single string for strings
2. normalizing the value (numeric: make it undefined, string: make it lower case, wrap in array if needed

However, on line 207 here we set the normalizedFieldValue to `''` when the numeric value needs to be `undefined` in order to have an allowedValue in the `validateFieldValueWithExceptions` function.

As I read it, this won't work. Is there a step I am missing or does this need to be updated slightly?

src/submission/exceptions/exceptions.ts
```diff
@@ -161,52 +189,56 @@ export const checkForProgramAndEntityExceptions = async ({
...
-    // Field value is a type we cannot allow exceptions for, validate as normal
-    if (!isAllowedTypeForException(fieldValue)) {
-      filteredErrors.push(validationError);
-      return;
-    }
+    const fieldSchema = entitySchema?.fields.find(fieldFilter(validationErrorFieldName));
+    const valueType = fieldSchema?.valueType;
+
+    const validNumericExceptionValue =
+      (valueType === 'number' || valueType === 'integer') &&
```
joneubank, line 204: Can reuse the `isNumericType` function I suggested in the other comment.

joneubank (commented):
Looks alright in terms of structure, but one section seems logically wrong to me just from reading the code. Maybe you can explain it or fix that?

src/submission/exceptions/exceptions.ts
demariadaniel: See line 219 -- for numeric exceptions, we just use what was submitted
We initialize `normalizedFieldValue` as a string b/c a) exceptions were originally for string fields and b) only string fields need to be normalized
For numeric values, we aren't doing any normalization

demariadaniel: So connecting the dots with your other comments -- if it's a numeric field and the record value is `''` we should be normalizing to undefined. I'll update.

src/submission/exceptions/exceptions.ts
demariadaniel: Updated

src/submission/exceptions/exceptions.ts
```diff
@@ -161,52 +194,55 @@ export const checkForProgramAndEntityExceptions = async ({
...
+    if (valueType === 'string' && isValidStringExceptionType(fieldValue)) {
+      // get normalized value for record, from either the string value or from the single string inside of the array.
+      // we should know from `isAllowedTypeForException` that the fieldValue is one of those two types.
+      const stringFieldValue = isSingleString(fieldValue) ? fieldValue[0] : fieldValue;
+      const normalizedString = normalizeExceptionValue(stringFieldValue);
+      normalizedValue = isSingleString(fieldValue) ? [normalizedString] : normalizedString;
+
+      normalizedFieldValue = normalizedString;
+    } else if (validNumericExceptionValue) {
+      normalizedFieldValue = fieldValue === '' ? undefined : fieldValue;
```
joneubank, line 223: :+1: thanks for pointing this out!

joneubank (approved):

PR #1110, opened 2024-01-22
Create update endpoint

ciaranschutte:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/routes/exception.ts
```diff
@@ -19,32 +19,59 @@
...
+
+/**
+ * Missing entity exceptions
+ */
+const missingEntityExceptionRouter = express.Router({ mergeParams: true });
 
-router.get('/', wrapAsync(exceptionApi.getProgramException));
-router.get('/entity', wrapAsync(exceptionApi.getEntityException));
+// GET
+missingEntityExceptionRouter.get('/', (req, resp) => resp.send({ msg: 'not implemented yet...' }));
```
joneubank, line 62: Return 500 with message Not Implemented

src/routes/exception.ts
```diff
@@ -19,32 +19,59 @@
...
+const notFound = (req: Request, res: Response) => res.status(404).send();
 
-router.use(ExceptionErrorHandler);
+// Parent
+const exceptionRouter = express.Router({ mergeParams: true });
+exceptionRouter.use(
+	'/property-exception',
+	featureFlags.FEATURE_SUBMISSION_EXCEPTIONS_ENABLED ? propertyExceptionRouter : notFound,
+);
+exceptionRouter.use('/missing-entity-exception', missingEntityExceptionRouter);
```
joneubank, line 75: Exceptions are in place now and won't be disabled, let's remove the feature flag entirely. Takes away the need for `notFound` endpoints.

src/exception/missing-entity-exceptions/api.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Request, Response } from 'express';
+import { createOrUpdate } from './repo';
+
+const parseBoolString = (queryString: string) => {
+	if (queryString) {
+		return queryString.toLowerCase() === 'true' || false;
+	} else {
+		return false;
+	}
+};
```
joneubank, line 30: Move this somewhere reusable?

src/exception/missing-entity-exceptions/api.ts
```diff
@@ -0,0 +1,50 @@
...
+
+class MissingEntityExceptionController {
```
joneubank, line 32: No need for a class, we can just export the functions directly from this module.

src/exception/missing-entity-exceptions/api.ts
```diff
@@ -0,0 +1,50 @@
...
+	async createEntityException(req: Request, res: Response) {
+		const programId = req.params.programId;
+		const donorSubmitterIds = req.body.donorSubmitterIds;
+		const isDryRun = parseBoolString(req.query['dry-run']);
+
+		const result = await createOrUpdate({
+			programId,
+			donorSubmitterIds,
+			isDryRun,
+		});
```
joneubank, line 43: This method doesn't have any logic for checking the provided values, its just an overwrite.

The logic for our API is additive, adding the new donor IDs provided in the request.

So we need to:
1. read the existing exception
2. add the new donor IDs
3. filter to make each entry unique
4. then save that list as the exception.

Additionally, we want to communicate to the requester how many of the IDs were added and how many already existed, so we can do those checks on the existing exception data and construct a matching type for return:

```ts
type createResponse = {
  "donorsAdded": string[],
  "donorsAddedCount": number,
 
  "donorsUnchanged": string[],
  "donorsUnchangedCount": number
}
```

src/exception/missing-entity-exceptions/api.ts
```diff
@@ -0,0 +1,50 @@
...
+
+		const status = result.success ? 201 : 422;
```
joneubank, line 45: only known failure case for createOrUpdate is a server failure to communicate with mongo so this would return error 500.

```ts
if(!result.success) {
  return res.status(500).send({/* error response object, maybe use a `failure()` object like we did in the other exception endpoints */})
}
```

src/resources/swagger.yaml
```diff
@@ -842,6 +842,26 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/EntityExceptionDonorIds'
+
+      responses:
+        '201':
+          description: 'successful upload'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
```
joneubank, line 863: Statuses don't match the request.

Please create an entity for the return type.

joneubank (changes requested):

joneubank:
Responding to your description of changes to the endpoints, it is clear we will need to update them for the existing exceptions as well as define new paths for the missing-entity exceptions. How about this proposal for the paths to the different types of exceptions:


```
Property Exceptions, program level:
/exception/property/{programId}

PropertyExceptions, entity level:
/exception/property/{programId}/entity

Missing Entity Exceptions:
/exception/missing-entity/{programId}
```

src/routes/exception.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1110/commits/39551ffbda67286a4e8241af094246aeb583453e

src/exception/missing-entity-exceptions/api.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1110/commits/31affb29aa5cc93ed03208d02974a9d630e0b301

src/exception/missing-entity-exceptions/api.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1110/commits/a195997c9806bf81f3cda4ad1edd892d9629277f

src/exception/missing-entity-exceptions/api.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1110/commits/a195997c9806bf81f3cda4ad1edd892d9629277f

src/exception/missing-entity-exceptions/api.ts
ciaranschutte: leaving class structure for decorators

src/routes/exception.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1110/commits/b811828b1992ae8706a879be10675dec5f880dfd

src/routes/exception/exception.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import * as express from 'express';
+import propertyExceptionRouter from './property';
+import missingEntityExceptionRouter from './missing-entity';
+
+const exceptionRouter = express.Router({ mergeParams: true });
+exceptionRouter.use('/property', propertyExceptionRouter);
+exceptionRouter.use('/missing-entity', missingEntityExceptionRouter);
+
+export default exceptionRouter;
```
joneubank, line 28: Rename this file to `index.ts`

src/resources/swagger.yaml
```diff
@@ -1840,3 +1869,30 @@ components:
...
         ]
+
+    MissingEntityExceptionDonorIds:
+      type: object
+      properties:
+        submitter_donor_ids:
+          description: 'Array of strings, each of which is a submitter donor id to add to this program/s exception. This endpoint is additive.'
+          type: array
+
+    MissingEntityCreateResp:
```
joneubank, line 1880: Lets use full words for clarity (Response): `MissingEntityCreateResponse`

src/resources/swagger.yaml
```diff
@@ -1415,6 +1437,13 @@ components:
               value:
                 msg: 'string'
                 code: 'INVALID_FILE_NAME'
+    MissingEntityCreate:
+      description: 'successful upload'
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/MissingEntityCreateResp'
```
joneubank, line 1445: I don't think this needs to be defined outside of where it is used. Can simplify the swagger file a bit by just moving this in place under the response 200 at line 861.

src/resources/swagger.yaml
```diff
@@ -842,6 +842,28 @@ paths:
...
+        - $ref: '#/components/parameters/PathProgramId'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MissingEntityExceptionDonorIds'
+
+      responses:
+        '201':
```
joneubank, line 860: This is Create or Update so it may not actually create an entity. Let's just use Status 200

src/routes/exception/missing-entity.ts
```diff
@@ -0,0 +1,40 @@
...
+import missingEntityExceptionApi from '../../exception/missing-entity-exceptions/api';
+
+/**
+ * Missing entity exceptions
+ */
+const missingEntityExceptionRouter = express.Router({ mergeParams: true });
+
+// GET
+missingEntityExceptionRouter.get('/:programId', (req, resp) =>
+	resp.status(500).send({ msg: 'not implemented yet...' }),
```
joneubank, line 31: 👍

src/utils/request.ts
```diff
@@ -0,0 +1,26 @@
...
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+export const parseBoolString = (queryString: string) => {
```
joneubank, line 20: This is a case where I would explicitly declare the return type of the function. This is because when I am writing it I know I want it to always return a `boolean`, so by declaring the return type TS will ensure that this function always returns what is expected.

```suggestion
export const parseBoolString = (queryString: string): boolean => {
```

src/exception/missing-entity-exceptions/api.ts
```diff
@@ -0,0 +1,47 @@
...
+			programId,
+			newDonorIds: donorSubmitterIds,
+			isDryRun,
+		});
+
+		if (!result.success) {
+			const { message, errors } = result;
+			return res.status(500).send({ message, errors });
+		} else {
+			return res.status(201).send(result.exception);
```
joneubank, line 42: Commented in the swagger... I think this should be 200 instead of 201 since it is create or update.

src/exception/missing-entity-exceptions/service.ts
```diff
@@ -0,0 +1,72 @@
...
+import { createOrUpdate, getByProgramId } from './repo';
+
+type CreateResponse = {
+	donorsAdded: string[];
+	donorsAddedCount: number;
+	donorsUnchanged: string[];
+	donorsUnchangedCount: number;
+};
+
+export const create = async ({
```
joneubank, line 30: This would benefit from TSDoc block for future readers.

src/exception/missing-entity-exceptions/service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	if (missingEntityExceptionResult.success) {
+		const currentDonorIds = missingEntityExceptionResult.exception.donorSubmitterIds;
+
+		// return unique donor ids
+		const donorSubmitterIds = [...new Set([...currentDonorIds, ...newDonorIds])];
+
+		// calc new and updated ids
+		const donorsAdded = donorSubmitterIds.filter((id) => !currentDonorIds.includes(id));
+		const donorsUnchanged = donorSubmitterIds.filter((id) => currentDonorIds.includes(id));
```
joneubank, line 49: Looks like this logic is a tiny bit off, you are filtering from `donorSubmitterIds` that already has all of the `currentDonorIds` and `newDonorIds` added. I think these want to be:
```suggestion
		// calc new and updated ids
		const donorsAdded = currentDonorIds.filter((id) => !newDonorIds.includes(id));
		const donorsUnchanged = currentDonorIds.filter((id) => newDonorIds.includes(id));
```

joneubank (changes requested):
Bunch of little things, looks really close to ready.

joneubank (approved):

PR #1111, opened 2024-01-22
Core Complete Calculation logic using Missing Entity Exception

joneubank:
## Link to Issue
- https://github.com/icgc-argo/roadmap/issues/1092

## Description

Core completion calculation needs to take into consideration whether a donor has a MissingEntityException. If they do, then the Treatment and Follow-up entities are not required to be core complete.

The mechanism for calculating core completion stats has been standardized - there are no longer different calculation cases that behave differently based on overrides or completion state. Now, core completion calculation will always result in the same result for a given donor.

Core-Completion is triggered by:
1. Data registration
2. Data submission
3. Dictionary migration

This removes the patch endpoint that allows admin to "override" core calculation stats. The CompletionStats object has had this override property removed and all the calculation logic that used these overrides are removed. A migration is included to remove all overrides from teh database. It has been checked that this feature was never used in the current ARGO data set so this migration is added to ensure consistency, though we don't expect to have any actual cases to update.

A caching utility for MissingEntityExceptions was created as a performance consideration to help when fetching exceptions for many donors. Using this cache to access the MissingEntityException repo will prevent repeatedly fetching exceptions for the same program when you are fetching exceptions for a long list of donors.

- [x] Remove core completion override logic, including Admin patch endpoint to override core completion stats
- [x] Migration added to remove all references to  override
- [x] Update core-completion logic to always be consistent (no calculate if flags)
- [x] Centralize code that calls for core completion updates
- [x] Add missing-entity-exceptions into core completion logic

### Type of Change

- [ ] Bug
- [x] Refactor
- [x] New Feature
- [ ] Release Candidate

- [x] Check branch ~~(code change PRs go to `develop` not master)~~ (intentionally into feature branch)
- [ ] Manual testing

### DIFF
--- a/src/exception/missing-entity-exceptions/missing-entity-exception-cache.ts
+++ b/src/exception/missing-entity-exceptions/missing-entity-exception-cache.ts
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2024 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { DeepReadonly } from 'deep-freeze';
+import { Donor } from '../../clinical/clinical-entities';
+import * as missingEntityExceptionRepo from '../../exception/missing-entity-exceptions/repo';
+
+/**
+ * Creates a short lived caching object that will store missing entity exceptions hashed by programId.
+ * This provides access to the MissingEntityExceptions stored in the database with caching by programId
+ * so that many donors can be checked for exceptions without having to make repeated database queries.
+ *
+ * It is built for use with the `stat-calculator.ts` function `updateSingleDonorCompletionStats`.
+ *
+ * @example
+ * ```
+ * const cache = createMissingEntityExceptionCache();
+ *
+ * // will fetch exceptions for program, or return cached list.
+ * const programDonorsWithException = await cache.getProgramExceptionDonors(someProgramId);
+ *
+ * // boolean value indicating if a donor has an exception, checked against cache
+ * const donorHasMissingEntityException = await cache.donorHasException(donor);
+ * ```
+ *
+ * @returns
+ */
+export const createMissingEntityExceptionCache = () => {
+	const programExceptionData: Map<string, string[]> = new Map();
+
+	/**
+	 * Retrieve list of submitterDonorIds that have the MissingEntityException for a specific program
+	 * @param programId
+	 * @returns
+	 */
+	const getProgramExceptionDonors = async (programId: string): Promise<string[]> => {
+		const programData = programExceptionData.get(programId);
+		if (programData) {
+			return programData;
+		}
+
+		const programExceptionResult = await missingEntityExceptionRepo.getByProgramId(programId);
+		const programException = programExceptionResult.success
+			? programExceptionResult.exception.donorSubmitterIds
+			: [];
+
+		programExceptionData.set(programId, programException);
+
+		return programException;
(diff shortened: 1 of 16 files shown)


src/exception/missing-entity-exceptions/missing-entity-exception-cache.ts
```diff
@@ -0,0 +1,87 @@
...
+ * // will fetch exceptions for program, or return cached list.
+ * const programDonorsWithException = await cache.getProgramExceptionDonors(someProgramId);
+ *
+ * // boolean value indicating if a donor has an exception, checked against cache
+ * const donorHasMissingEntityException = await cache.donorHasException(donor);
+ * ```
+ *
+ * @returns
+ */
+export const createMissingEntityExceptionCache = () => {
```
joneubank, line 44: This is a tool for accessing the `MissingEntityException` repo data. Since the exceptions are stored per program, when we are checking for an exception for many donors we don't need to call the repo repeatedly for the same program. This will store the results read from the repo for a single program and allow that data to be reused throughout the lifetime of this cache.

src/schemas/gqlTypeDefs.ts
```diff
@@ -20,350 +20,350 @@
...
+	}
+
+	"""
+	Completion Data for a given Donor
+	"""
+	type CompletionStats {
+		coreCompletion: CoreCompletionFields
+		coreCompletionDate: String
+		coreCompletionPercentage: Float
+		hasMissingEntityException: Boolean
```
joneubank, line 187: This is the only change, updating the response structure based on the updated CompletionStats type.

src/submission/migration/migration-manager.ts
```diff
@@ -18,811 +18,804 @@
...
+		let migrationDone = false;
+		if (!migration._id) {
+			throw new Error('Migration should have an id');
+		}
+		const migrationId = migration._id;
+		const dryRun = migration.dryRun;
+		const breakingChangesEntitesCache: { [versions: string]: ClinicalEntitySchemaNames[] } = {};
+		const coreFieldChangesEntitiesCache: { [versions: string]: ClinicalEntitySchemaNames[] } = {};
+
+		const missingEntityExceptionCache = createMissingEntityExceptionCache();
```
joneubank, line 442: Create cache to use to fetch missing entity exceptions for each donor.

src/submission/migration/migration-manager.ts
```diff
@@ -18,811 +18,804 @@
...
+					});
+
+					invalidCount += 1;
+					continue;
+				} else {
+					if (!dryRun) {
+						const updatedDonor = await updateSingleDonorCompletionStats(
+							donor,
+							missingEntityExceptionCache,
+						);
```
joneubank, line 495: Get update donor info including core completion stats, function uses cache for fetching exception data.

src/submission/submission-to-clinical/merge-submission.ts
```diff
@@ -18,150 +18,170 @@
...
-          break;
-        case ClinicalTherapySchemaNames.find(tsn => tsn === entityType):
-          entityWithUpdatedInfo = updateOrAddTherapyInfoInDonor(donor, record, entityType, true);
-          break;
-        default:
-          throw new Error(`Entity ${entityType} not implemented yet`);
-      }
-
-      // update clinical entity stats
-      updateDonorStatsFromSubmissionCommit(donor, entityType);
```
joneubank, line 107: This has been removed. The merge-submission code no longer recalculates core completion. This is done separately by the code that is submitting the data. merge-submission now is only updating the clinical entity data in the donor document.

src/submission/submission-to-clinical/submission-to-clinical.ts
```diff
@@ -60,117 +56,102 @@ import * as messenger from '../submission-updates-messenger';
...
-    // write each updated donor to the db
-    await donorDao.updateAll(verifiedDonorDTOs.map(dto => F(dto)));
-
-    // If the save completed without error, we can delete the active registration
-    submissionRepository.deleteByProgramId(activeSubmission.programId);
-
-    sendMessageOnUpdatesFromClinicalSubmission(activeSubmission);
-  } catch (err) {
-    throw new Error(`Failure occured saving clinical data: ${err}`);
-  }
```
joneubank, line 173: This donor validation logic was moved to a separate function that will return the list of donors with the validation data updated.

The completion stat check was pulled out of this section and is applied to the array of donors in a separate step.

src/submission/submission-to-clinical/submission-to-clinical.ts
```diff
@@ -60,117 +56,102 @@ import * as messenger from '../submission-updates-messenger';
...
+		);
+	}
+
+	// Update donors with submission data
+	//  1. merge submission data into donor documents
+	//  2. validate updated donors vs current schema
+	//  3. recalculate donors core complete stats
+	const mergedDonors = await mergeActiveSubmissionWithDonors(activeSubmission, donorDTOs);
+	const donorsWithSchemaValidation = await validateDonorsWithCurrentDictionary(mergedDonors);
+	const updatedDonors = await updateDonorsCompletionStats(donorsWithSchemaValidation);
```
joneubank, line 142: Here we can see the donor validation and core completion updates happening in separate steps to the entire donor array instead of being applied individually to each donor.

src/utils/results.ts
```diff
@@ -4,7 +4,7 @@ export type Failure = { success: false; message: string; errors?: any };
 export type Result<T> = Success<T> | Failure;
 
 // helpers
-export const success = <T>(data: T): Success<T> => ({ success: true, exception: data });
+export const success = <T>(exception: T): Success<T> => ({ success: true, exception });
```
joneubank, line 7: Naming consistency. We want to update this whole type to use `data` instead of `exception`, but we will make these changes at the end following all the merges (fewer conflicts hopefully).

src/submission/submission-to-clinical/stat-calculator.ts
```diff
@@ -76,141 +96,125 @@ const getEmptyCoreStats = (): CompletionStats => ({
...
-	// Intentionally create a mutable Donor copy for coreCompletion updates
-	const donorCopy = cloneDeep(donor) as Donor;
-
-	let newCompletionStats = donorCopy.completionStats
-		? { ...donorCopy.completionStats }
+export const calculateDonorCoreCompletionStats = (
+	donor: Donor,
+	hasMissingEntityException: boolean,
+): CompletionStats => {
+	let newCompletionStats: CompletionStats = donor.completionStats
```
ciaranschutte, line 115: this doesn't need to `let`
`const` is memory of object itself, not properties of object

ciaranschutte (approved):
looks great. thanks for all the comments and clear code.

src/submission/submission-to-clinical/stat-calculator.ts
joneubank: right you are.

PR #1113, opened 2024-01-23
Feat/#1095 get single program entity exception

ciaranschutte:
## Link to Issue
https://github.com/icgc-argo/roadmap/issues/1092

### Type of Change

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/exception/missing-entity-exceptions/api.ts
```diff
@@ -22,9 +22,20 @@ import { Request, Response } from 'express';
...
+import { getByProgramId, listAll } from './repo';
 
 class MissingEntityExceptionController {
+	@HasFullReadAccess()
+	async getProgramException(req: Request, res: Response) {
+		const result = await getByProgramId(req.params.programId);
+		if (!result.success) {
+			const { message, errors } = result;
+			return res.status(500).send({ message, errors });
+		} else {
```
joneubank, line 34: When we have no database entry, can we instead reply with a blank exception object?
example:
```ts
{
  programId: "SomeId",
  donorSubmitterIds: []
}
```

joneubank: Also note for reference... if the response shows no entry for the parameters the user requested, thats a 4xx error of some sort, based on user inputs, not a 5xx server error.

joneubank (changes requested):
Requesting change to response when we have no DB record for a program ID.

src/exception/missing-entity-exceptions/api.ts
ciaranschutte: If there's no db entry that's happening already https://github.com/icgc-argo/argo-clinical/blob/feat%2Fmissing-entity-exception/src/exception/missing-entity-exceptions/repo.ts#L84

ciaranschutte: in this case though it is sent with status 200 and the object you suggested above

joneubank (approved):

src/exception/missing-entity-exceptions/api.ts
joneubank: Yes that is correct lol. Already handled.

PR #1114, opened 2024-01-23
Feat/#1096 remove donors from program

ciaranschutte:
## Link to Issue
https://github.com/icgc-argo/roadmap/issues/1092

## Description

remove donors from program

### Type of Change

- [ ] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


joneubank (approved):
This exactly matches the update request which is great, works perfect.

There is one particular check that is not being done here and that is to ensure that we don't remove exceptions on donors that have used the exception to become core-complete.

I think we can merge this and add that in as an additional check after.

src/routes/exception/exception.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import * as express from 'express';
+import propertyExceptionRouter from './property';
+import missingEntityExceptionRouter from './missing-entity';
+
+const exceptionRouter = express.Router({ mergeParams: true });
+exceptionRouter.use('/property', propertyExceptionRouter);
+exceptionRouter.use('/missing-entity', missingEntityExceptionRouter);
+
+export default exceptionRouter;
```
demariadaniel, line 28: Curious, why a separate folder + file for these routes, separate from the rest of the exception routes at `/routes/exception.ts` ?

demariadaniel: OK I see ... separation into property router vs missing-entity router

demariadaniel (approved):

src/routes/exception/exception.ts
ciaranschutte: just organisation. 

the previous exceptions are now "exceptions/property" and these newly added missing entity exception ones are under "exceptions/missing-entity".

containing all exception logic in a single folder & under a single route declaration in app means we could add/remove/extend functionality easier. Example: what if all exception endpoints needed an extra auth check? We could just add a middleware at the `exception` route instead of in two places.

I have noticed there is a duplicate file here. should only be index and not exception for the module exporting.

PR #1139, opened 2024-02-16
🕷️ 1115 fix clinical data resolver completion

demariadaniel:
## Link to Issue

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/1115

## Description

Fixes several bugs related to ClinicalData resolver:
- CompletionState is now populated correctly
- Default value for query sort field
- Error handling when ClinicalErrors migration schema is not found
- API Input filter type updated

### Type of Change

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


src/schemas/clinical-resolvers/clinicalData.ts
```diff
@@ -75,8 +76,12 @@ const convertClinicalDataToGql = (
 
 const clinicalDataResolver = async (obj: unknown, args: ClinicalDataVariables) => {
 	const { programShortName, filters } = args;
+	const { completionState: state, sort = 'donorId' } = filters;
+	const completionState = completionFilters[state];
```
ciaranschutte, line 80: 👍

ciaranschutte: great fix - thank you

src/clinical/api/types.ts
```diff
@@ -15,6 +16,8 @@ export const ClinicalDataApiBody = zod.object({
 });
 export type ClinicalDataApiBody = zod.infer<typeof ClinicalDataApiBody>;
 
+export type ClinicalDataPaginatedBody = ClinicalDataApiBody & PaginationQuery;
```
ciaranschutte, line 19: I think we can avoid a messy codebase here as duplication of types is popping up.

https://github.com/icgc-argo/argo-clinical/blob/develop/src/clinical/clinical-service.ts#L72
This seems to be typing the same thing.

`PaginatedClinicalQuery` is still used throughout and

src/clinical/clinical-service.ts
```diff
@@ -425,7 +426,11 @@ export const getValidRecordsPostSubmission = async (
 
 		const migrationDictionary = await dictionaryManager
 			.instance()
-			.loadSchemaByVersion(schemaName, migrationVersion);
+			.loadSchemaByVersion(schemaName, migrationVersion)
+			.catch(async (err) => {
+				L.error('getValidRecordsPostSubmission error finding migration schema', err);
+				return await dictionaryManager.instance().getCurrent();
+			});
```
ciaranschutte, line 433: If there's an error loading a specific version of the dictionary then we default to the current?
This seems like new functionality/breaking change. cc @joneubank for 2nd pair of eyes

ciaranschutte (commented):
@demariadaniel one q for @joneubank before I approve 
(one comment re types but not a blocker)

ciaranschutte:
Something I'm trying to get better at is 1 issue/ 1 PR. Difficult esp when there's "quick fixes" - but I think it helps us avoid the situations we sometimes get into where we're held up with PRs

src/clinical/clinical-service.ts
demariadaniel: This is due to an edge case I'm encountering on localhost; arguably should be a separate PR, but it could cause further Submitted Data issues (has slowed me down) which is why I included it here

When the requested migration dictionary is not found, this breaks the Clinical Data query. So I've added `getCurrent` as a fallback.
This is a problem with my local DB getting out of sync w/ Lectern Dev somehow; however it's also an edge case I'd hate to see in Prod or Staging

src/clinical/api/types.ts
demariadaniel: Yes this crossed my mind as well, let me review

demariadaniel: OK, see: https://github.com/icgc-argo/argo-clinical/pull/1139/commits/3cd644374e8a16fff765273fdfc81f52b8d1ca80

I reverted to using only `PaginatedClinicalQuery`, and it turns out `programShortName` was unnecessary to include in this type at all, b/c every function that uses this query is providing programId as the first argument, and query only concerns other details like donorId, entityType, and pagination

ciaranschutte:
Screenshot 2024-02-20 at 11 39 47 AM
@demariadaniel  missing type update here: `src/submission/validation-clinical/therapy.ts`
and the completion state has a type error also

`npm run type-check` should get you the same result as above

build won't work without TS checks passing

demariadaniel:
> Screenshot 2024-02-20 at 11 39 47 AM @demariadaniel missing type update here: `src/submission/validation-clinical/therapy.ts` and the completion state has a type error also
> 
> `npm run type-check` should get you the same result as above
> 
> build won't work without TS checks passing

Sorry, app ran and vscode showed no errors so missed this.
The issue w/ Type duplication is because incoming API filters and Mongo filters are different, but very similar
Will untangle in this commit

ciaranschutte:
> > Screenshot 2024-02-20 at 11 39 47 AM @demariadaniel missing type update here: `src/submission/validation-clinical/therapy.ts` and the completion state has a type error also
> > `npm run type-check` should get you the same result as above
> > build won't work without TS checks passing
> 
> Sorry, app ran and vscode showed no errors so missed this. The issue w/ Type duplication is because incoming API filters and Mongo filters are different, but very similar Will untangle in this commit

np. I thought we had github build checks for this repo that should have caught the TS thing, will need to investigate that.

demariadaniel:
> np. I thought we had github build checks for this repo that should have caught the TS thing, will need to investigate that.

@ciaranschutte Updated
https://github.com/icgc-argo/argo-clinical/pull/1139/commits/e3587ebd5b024d9ad7f01c61882dc991119b805f

src/clinical/api/types.ts
```diff
@@ -1,4 +1,5 @@
 import { EntityAlias } from '../../common-model/entities';
+import { PaginationQuery } from '../clinical-service';
```
justincorrigible, line 2: Looks like an unused import?

src/clinical/api/clinical-api.ts
```diff
@@ -243,12 +241,13 @@ class ClinicalController {
 		const completionState = completionFilters[state] || {};
 
 		// FE filters digits out of search text for Donor search
-		const query: ClinicalDonorEntityQuery = {
-			programShortName,
+		const query: ClinicalDataDbQuery = {
```
justincorrigible, line 244: "Clinical Data Database Query" -> "Clinical Database Query" to avoid redundancy?

src/clinical/clinical-service.ts
```diff
@@ -69,19 +68,21 @@ export type PaginationQuery = {
 	sort: string;
 };
 
-export type PaginatedClinicalQuery = ClinicalDonorEntityQuery & PaginationQuery;
+type ClinicalDataPaginatedQuery = ClinicalDonorEntityQuery & PaginationQuery;
+
+export type ClinicalDataDbQuery = ClinicalDataPaginatedQuery & {
```
justincorrigible, line 73: "Clinical Data Database Query" -> "Clinical Database Query" to avoid redundancy?

src/clinical/api/types.ts
demariadaniel: Snuck past me, removed

ciaranschutte (approved):

src/clinical/clinical-service.ts
joneubank: Sorry to respond so late after this is merged. I am surprised a data query is reliant on fetching a dictionary from Lectern. All dictionaries that we use, should be already loaded into the clinical DB - is that not the case in this instance? This shouldn't be dependent on fetches out to lectern, just read the relevant dictionary from the DB.

demariadaniel: Right we did chat about this -- I will make a quick PR to revise. For now this unblocks local dev using Submitted Data.

Issue #1125, opened 2024-02-01
🐛 Surgery table is not able to be updated


lindaxiang:
## Describe the bug
The surgery table is not able to be updated once submitted. 

## Steps To Reproduce

Steps to reproduce the behaviour:
1. Go to 'https://platform-ui.argo-qa.cumulus.genomeinformatics.org/'
2. Click on project 'PACA-CA' or other projects
3. If you already have surgery records submitted
4. Re-submit the surgery table with or without changes
5. Click the `Validate submission` button
6. it will give the error message:
`
The submitter_specimen_id '*****' has already been associated with a surgery in the current or previous submission. Specimen can only be submitted once for a single surgery.
`
## Expected behaviour
The surgery table should be updated accordingly without errors.

demariadaniel:
Bug is actually the result of existing validation created in: 
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/argo-clinical/773

&

https://github.com/icgc-argo/argo-clinical/issues/738

Will need to discuss solution that allows updating existing records, without creating duplicates
Any current issues with existing records can be resolved by deleting records & resubmitting

demariadaniel:
Tested successful update scenario:

<img width="836" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL0xPQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--21052b00bdc2afffd4495fbd221d5db4252f8fde/Screenshot%202024-03-11%20at%205.03.04%E2%80%AFPM.png" alt="Screenshot 2024-03-11 at 5.03.04 PM.png" />

demariadaniel:
Update success case TSV file:
[surgery_resubmit_test1_update.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1hPQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--58d65c80ef0ddb6f15898c3ceb371114e2fff083/surgery_resubmit_test1_update.txt)

(tsv is not supported by zenhub, converted to txt)

demariadaniel:
Validation successfully shows error for duplicate record:

[...]

PR #1162, opened 2024-03-11
🪰 1125 Allow Updates to Surgery Table

demariadaniel:
## Link to Issue

Testing output & records here:
#1125 

## Description

Allows submitting updates to existing surgery records
Throws errors when Submission duplicates an existing record, or when multiple records are submitted (existing behavior)

### Type of Change

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing


test/unit/submission/validation.spec.ts
```diff
@@ -23,3207 +23,3208 @@ import _ from 'lodash';
...
+		});
+
+		it('should error when multiple surgeries with the same submitter_specimen_id are found', async () => {
+			const existingDonor = stubs.validation.existingDonor09();
+			const submissionRecordsMap = {};
+			ClinicalSubmissionRecordsOperations.addRecord(
+				ClinicalEntitySchemaNames.SURGERY,
+				submissionRecordsMap,
+				{
+					program_id: 'TEST-CA',
```
demariadaniel, line 2918: program_id added to make records duplicate eachother
Majority of changes in this file are just linting

demariadaniel: PR to Lint: https://github.com/icgc-argo/argo-clinical/pull/1163

test/unit/submission/validation.spec.ts
```diff
@@ -23,3207 +23,3245 @@ import _ from 'lodash';
...
+					donorSubmitterId: 'ICGC_0002',
+					value: 'ICGC_0002',
+				},
+			};
+
+			chai.expect(result[ClinicalEntitySchemaNames.SURGERY].dataErrors.length).equal(1);
+			chai.expect(result[ClinicalEntitySchemaNames.SURGERY].dataErrors[0]).to.deep.eq(error);
+		});
+
+		it('should error when multiple surgeries with the same submitter_specimen_id are found', async () => {
```
demariadaniel, line 2948: Previous test, updated for new behaviour

src/submission/validation-clinical/surgery.ts
```diff
@@ -80,20 +81,8 @@ export const validate = async (
...
-					DataValidationErrors.DUPLICATE_SUBMITTER_SPECIMEN_ID_IN_SURGERY,
-					DonorFieldsEnum.submitter_donor_id,
-					{
-						submitter_specimen_id: therapyRecord[SurgeryFieldsEnum.submitter_specimen_id],
-					},
-				),
-			);
+			checkSurgeryDuplicateOrUpdate(therapyRecord, surgeries, existingSurgeryInDB, errors);
 		} else {
 			validateSurgeryByDonorAndTreatment(therapyRecord, existentDonor, mergedDonor, errors);
```
joneubank, line 87: Are these two functions (`checkSurgeryDuplicateOrUpdate` and `validateSurgeryByDonorAndTreatment`) appending items to the `errors` array?

In general we don't want to have functions mutating variables from outside their scope. Better to have these functions return the new items that need to be added to the array and then we can append the new errors. I am writing this comment before reading the internals of your new function, I assume you're simply using the same interface pattern as the existing function - when I see the errors array being passed into this function I would make an assumption that the logic of this method is dependent on the content of this array... so I'll ask naively, does this function care about the existing content of the `errors` array?

I know the `validateSurgeryByDonorAndTreatment` is existing, not added here, but if all of the above is valid and we do want to modify the new function to return errors, it would be best to modify both in the same way (consistent patterns). This way we can improve the code base to be better than when we started.

src/submission/validation-clinical/surgery.ts
```diff
@@ -199,3 +188,40 @@ function checkSurgeryTypeEquality(
...
+			utils.buildSubmissionError(
+				therapyRecord,
+				DataValidationErrors.DUPLICATE_SUBMITTER_SPECIMEN_ID_IN_SURGERY,
+				DonorFieldsEnum.submitter_donor_id,
+				{
+					submitter_specimen_id: therapyRecord[SurgeryFieldsEnum.submitter_specimen_id],
+				},
+			),
+		);
+	}
```
joneubank, line 226: Related to previous comment... this is the only interaction with the `errors` array, can we instead have this function return the new items instead of mutating a variable that is passed in?

src/submission/validation-clinical/surgery.ts
```diff
@@ -199,3 +188,40 @@ function checkSurgeryTypeEquality(
...
+
+// Prevents submitting duplicate records; allows updating existing records with new values
+function checkSurgeryDuplicateOrUpdate(
+	therapyRecord: DeepReadonly<SubmittedClinicalRecord>,
+	prevSurgeries: DeepReadonly<Therapy>[],
+	existingSurgery: DeepReadonly<Therapy> | undefined,
+	errors: SubmissionValidationError[],
+) {
+	// Clone Submitted Record, minus index key, which is not stored on Clinical Records
+	const submissionClone = (({ index, ...rest }) => ({ ...rest }))(therapyRecord);
```
joneubank, line 200: This line is clever enough to need its own comment 😛 . I think it works as expected but is there a simpler way, or can we use an `omit` function? This repo already has lodash so just using [`_.omit`](https://lodash.com/docs/2.4.2#omit) would be more readable.

src/submission/validation-clinical/surgery.ts
```diff
@@ -199,3 +188,40 @@ function checkSurgeryTypeEquality(
...
+
+	// Determine if Submission is duplicating existing record, or if Submission is an update
+	// Sort insures [A,B,C] is not compared with [A,C,B]
+	const submissionValues = Object.values(submissionClone).sort();
+	const existingRecordValues = existingSurgery
+		? Object.values(existingSurgery.clinicalInfo).sort()
+		: [];
+	const submissionIsDuplicate =
+		submissionValues.length === existingRecordValues.length &&
+		isEqual(submissionValues, existingRecordValues);
```
joneubank, line 210: Members of `submissionValues` could be arrays, these arrays will not be found as equal since they are compared by reference value. Therefore this `isEqual` will not work in cases with therapies that have array properties. This may not be the case in the current dictionary, but for future proofing we will need a recursive equal check.

joneubank (changes requested):

src/submission/validation-clinical/surgery.ts
joneubank: Even simpler approach:

```ts
const { index, ...submissionClone } = therapyRecord;
```

src/submission/validation-clinical/surgery.ts
demariadaniel: I've tested this with array values and it worked but will give this a second look

src/submission/validation-clinical/surgery.ts
demariadaniel: Destructuring index leaves an unused variable, so I'll go with omit, should have used that in the first place
Updated

src/submission/validation-clinical/surgery.ts
demariadaniel: Yeah this is a legacy pattern used across all the validation files, I've been hesitant to refactor b/c it's widespread

src/submission/validation-clinical/surgery.ts
demariadaniel: Updated here:
https://github.com/icgc-argo/argo-clinical/pull/1162/commits/9a31dad1404f4c4dcc5e296063d71b825e930539

Could also use a `let errors; errors = [...errors, errorFunction()]` approach using reassignment
Half dozen of one, six of the other 🤷

src/submission/validation-clinical/surgery.ts
demariadaniel: From the Docs: "This method supports comparing arrays"
https://lodash.com/docs/4.17.15#isEqual

I added array values to the unit test for validation

src/submission/validation-clinical/surgery.ts
```diff
@@ -82,9 +82,19 @@ export const validate = async (
 		) as DeepReadonly<Therapy>;
 
 		if (existingSurgeryInDB || surgeries.length > 1) {
-			checkSurgeryDuplicateOrUpdate(therapyRecord, surgeries, existingSurgeryInDB, errors);
+			const duplicationError = checkSurgeryDuplicateOrUpdate(
+				therapyRecord,
+				surgeries,
+				existingSurgeryInDB,
+			);
+			if (duplicationError) errors.push(duplicationError);
```
joneubank, line 90: We want to always use brackets after an if statement.

- It keeps every logical block consistent which will help with the reader's ability to spot the start and end of the code block.
- It prevents future errors from editing mistakes. For example, where the single statement that is not wrapped in brackets is removed accidentally, making the next statement in the code now dependent on the code control statement before it.

src/submission/validation-clinical/surgery.ts
```diff
@@ -82,9 +82,19 @@ export const validate = async (
...
 		} else {
-			validateSurgeryByDonorAndTreatment(therapyRecord, existentDonor, mergedDonor, errors);
+			const surgeryNotEqualError = validateSurgeryByDonorAndTreatment(
+				therapyRecord,
+				existentDonor,
+				mergedDonor,
+			);
+			if (surgeryNotEqualError) errors.push(surgeryNotEqualError);
```
joneubank, line 97: Brackets after if statement.

src/submission/validation-clinical/surgery.ts
```diff
@@ -194,10 +201,9 @@ function checkSurgeryDuplicateOrUpdate(
 	therapyRecord: DeepReadonly<SubmittedClinicalRecord>,
 	prevSurgeries: DeepReadonly<Therapy>[],
 	existingSurgery: DeepReadonly<Therapy> | undefined,
-	errors: SubmissionValidationError[],
 ) {
 	// Clone Submitted Record, minus index key, which is not stored on Clinical Records
-	const submissionClone = (({ index, ...rest }) => ({ ...rest }))(therapyRecord);
+	const submissionClone = omit(therapyRecord, 'index');
```
joneubank, line 206: 👍

src/submission/validation-clinical/surgery.ts
```diff
@@ -117,7 +127,6 @@ function validateSurgeryByDonorAndTreatment(
 	therapyRecord: DeepReadonly<SubmittedClinicalRecord>,
 	existentDonor: DeepReadonly<Donor>,
 	mergedDonor: Donor,
-	errors: SubmissionValidationError[],
```
joneubank, line 131: 👍

joneubank (changes requested):
Quick changes needed, new updates are good.

src/submission/validation-clinical/surgery.ts
demariadaniel: Updated

joneubank (approved):

PR #1176, opened 2024-04-15
Clinical id service integration

### LINKED ISSUE #1094: Integrate ID generation with the Clinical Service.

UmmulkiramR:
Donor Repository methods have been modified to fetch the entity ids from the ID service before saving the donor in the collection.
ID generator is the one making the calls to the ID service.


src/clinical/donor-repo.ts
```diff
@@ -413,27 +416,44 @@ export const donorDao: DonorRepository = {
...
 		return F(MongooseUtils.toPojo(newDonor) as Donor);
 	},
 
-	async updateAll(donors: DeepReadonly<Donor>[]) {
-		const newDonors = donors.map((donor) => {
-			const newDonor = new DonorModel(donor);
+	async updateAll(donors: Donor[]) {
+		console.log('donors array in updateAll2: ' + donors[0].submitterId);
+		const newDonors = donors.map(async (donor) => {
+			// await someFunction();
```
demariadaniel, line 431: Unused commented code to remove

src/clinical/donor-repo.ts
```diff
@@ -413,27 +416,44 @@ export const donorDao: DonorRepository = {
...
+		);
+		const result = await Promise.all(
+			newDonors.map((donor) =>
+				donor.then((d) => {
+					return F(MongooseUtils.toPojo(d) as Donor);
+				}),
+			),
+		);
+
+		return result; // newDonors.map((donor) =>  donor.then((d) => { return F(MongooseUtils.toPojo(d) as Donor);}).then(res => res));
```
demariadaniel, line 452: Here as well

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+	const token = await getToken();
+	const headers = {
+		headers: {
+			Authorization: 'Bearer ' + token,
+		},
+	};
+
+	try {
+		const response = await axios.get(
+			// `http://localhost:9001/${req.programId}/${req.submitterId}/${req.submitterDonorId}/${req.testInterval}/${req.family_relative_id}/${req.comorbidityTypeCode}/${req.entityType}`,
```
demariadaniel, line 94: Unused

src/decorators/index.ts
```diff
@@ -123,11 +123,13 @@ const scopeCheckGenerator = (
 				scopesGenerator(programId),
 				request,
 				response,
-			);
-			if (unauthorizedResponse !== undefined) {
+			); // REMOVE THIS
```
demariadaniel, line 126: Seems pretty clear what to do here 😄

src/clinical/donor-repo.ts
```diff
@@ -413,27 +416,44 @@ export const donorDao: DonorRepository = {
...
 	},
 
-	async update(donor: DeepReadonly<Donor>) {
-		const newDonor = new DonorModel(donor);
+	async update(donor: Donor) {
+		const dnr = await setEntityIds(donor);
+		const newDonor = new DonorModel(await setEntityIds(donor));
 		unsetIsNewFlagForUpdate(newDonor);
-
+		console.log('newDonor._id' + newDonor._id);
```
demariadaniel, line 423: There are a lot of console.logs added and I'm not sure which are intentional and which are left over from testing

It would be good to replace those you want to keep with the `L` logger used elsewhere in the app; do a search for `const L` or `L.info` for examples

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+	programId: 'null',
+	submitterId: 'null',
+	submitterDonorId: 'null',
+	testInterval: 'null',
+	family_relative_id: 'null',
+	comorbidityTypeCode: 'null',
+	entityType: 'null',
+};
+
+export interface PartialDonor {
```
demariadaniel, line 60: This basically duplicates type `Donor`, could we achieve the same thing doing `interface PartialDonor extends Partial<Donor> {}` or something similar?

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+	// -- DONOR --
+	const donorId = await getId({
+		...request,
+		programId: donor.programId as string,
+		submitterId: donor.submitterId as string,
+		submitterDonorId: donor.submitterId as string,
+		entityType: ClinicalEntitySchemaNames.DONOR,
+	});
+	donor.donorId = donorId;
+	console.log(donor.donorId);
```
demariadaniel, line 124: This should print more than just a number, it needs context, at the very least: 'donor.donorId: ` similar to the other logs in this function

src/submission/submission-api.ts
```diff
@@ -313,7 +313,7 @@ class SubmissionController {
 
 	// GQL Query Methods
 	async commitActiveSubmissionData(programId: string, egoToken: string, versionId: string) {
-		queryHasProgramWriteAccess(programId, egoToken);
+		// queryHasProgramWriteAccess(programId, egoToken); // REMOVE THIS
```
demariadaniel, line 316: As well ^

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+		},
+	};
+
+	try {
+		const response = await axios.get(
+			// `http://localhost:9001/${req.programId}/${req.submitterId}/${req.submitterDonorId}/${req.testInterval}/${req.family_relative_id}/${req.comorbidityTypeCode}/${req.entityType}`,
+			config.getConfig().tokenUrl() +
+				`${req.programId}/${req.submitterId}/${req.submitterDonorId}/${req.testInterval}/${req.family_relative_id}/${req.comorbidityTypeCode}/${req.entityType}`,
+			headers,
+		);
```
demariadaniel, line 98: I would typically define this url in a variable which is then passed to `axios.get` rather than putting logic in the function arguments

```
const url = `${config.getConfig().tokenUrl()}/${req.programId}/${req.submitterId}/${req.submitterDonorId}/${req.testInterval}/${req.family_relative_id}/${req.comorbidityTypeCode}/${req.entityType}`;

const response = await axios.get(url, headers);
```

Also note I added a trailing slash between `tokenUrl` and the string of paths `req.programId/` etc ... if youre defining the tokenUrl with a slash then this is not needed

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+		console.log('getId response: ' + response.data.entityId + ' - ' + response.data.entityType);
```
demariadaniel, line 99: Just FYI string concatenation like this is generally out of favour compared to template literals
```
`getId response: ${response.data.entityId} - ${response.data.entityType}`
```

String + value (concatenation) can sometimes have unexpected results when parsing a value that is not originally a string
Template literals are also _slightly_ cleaner to read + write

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+				'null';
+
+			const id = await getId({
+				...request,
+				programId: donor.programId as string,
+				submitterId: submitterBiomarkerId,
+				submitterDonorId: bm.clinicalInfo.submitter_donor_id as string,
+				testInterval: bm.clinicalInfo.test_interval as string,
+				entityType: ClinicalEntitySchemaNames.BIOMARKER,
+			});
```
demariadaniel, line 174: When grabbing many values from the same place you can use destructuring variable assignment to simplify things:

```
			const {
				submitter_donor_id,
				submitter_specimen_id,
				submitter_primary_diagnosis_id,
				submitter_follow_up_id,
				submitter_treatment_id,
				test_interval,
			} = bm.clinicalInfo;

			const submitterBiomarkerId =
				submitter_specimen_id?.toString() ??
				submitter_primary_diagnosis_id?.toString() ??
				submitter_follow_up_id?.toString() ??
				submitter_treatment_id?.toString() ??
				'null';

			const id = await getId({
				...request,
				programId: donor.programId as string,
				submitterId: submitterBiomarkerId,
				submitterDonorId: submitter_donor_id as string,
				testInterval: test_interval as string,
				entityType: ClinicalEntitySchemaNames.BIOMARKER,
			});
```

demariadaniel (changes requested):
Functionality looks good but there's some clean up to do and some style recommendations before merging

src/clinical/id-generator.ts
demariadaniel: Also generally prefer descriptive variable names -- `biomarker` vs `bm`

ciaranschutte:
@demariadaniel @UmmulkiramR 
```js
!!'null' // true
!!null // false
```

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,293 @@
...
+	submitterId: any;
+	submitterDonorId: string;
+	testInterval: string;
+	family_relative_id: string;
+	comorbidityTypeCode: string;
+	entityType: string;
+}
+
+const request = {
+	programId: 'null',
```
justincorrigible, line 51: As discussed over zoom, this was made as a conscious choice, WIP in the context of query params in http requests, geared towards representing the difference between empty values for a param (null), vs missing parameters altogether (undefined). Will be addressed in this PR.

UmmulkiramR: have replaced the the null with a '-'

src/clinical/id-generator.ts
UmmulkiramR: that's a good idea. Thanks.

src/clinical/id-generator.ts
UmmulkiramR: done. Thanks.

src/clinical/id-generator.ts
UmmulkiramR: have removed all unwanted console logs

src/clinical/donor-repo.ts
UmmulkiramR: done. removed.

src/clinical/donor-repo.ts
UmmulkiramR: removed all console logs.

src/clinical/id-generator.ts
UmmulkiramR: I have changed bm to biomarker and for all other entities. But I'd prefer not to use the destructured variable assignment here for now

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,263 @@
...
+
+export async function setEntityIdsForDonors(donors: Donor[]) {
+	const donorsWithIds = donors.map(async (donor) => {
+		return await setEntityIds(donor);
+	});
+	return await Promise.all(donorsWithIds);
+}
+
+export async function setEntityIds(donor: PartialDonor) {
+	const submitterDonorId = donor.submitterId as string;
```
demariadaniel, line 87: One more thing I want to take a look at, I think we can update PartialDonor like this:

```
export interface PartialDonor extends Partial<Donor> {
	submitterId: string;
	programId: string;
}
```

and then you can remove repeated use of `as string` in this function and elsewhere. These two keys are typically required keys in `Donor` so I think it fits your usage.

We try to avoid using `as` because it can create situations where you're misleading Typescript. For example, `donor.submitterId` could actually be `undefined` and TS wouldn't know.

Tagging @justincorrigible (may want to ask Jon?) because `as` usage is something we've discussed frequently before so thought I'd get a second opinion

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+	submitterFollowUpId: '-',
+	submitterTreatmentId: '-',
+	submitterDonorId: '-',
+	testInterval: '-',
+	family_relative_id: '-',
+	comorbidityTypeCode: '-',
+	entityType: '-',
+};
+
+export interface PartialDonor extends Partial<Donor> {}
```
joneubank, line 56: Doesn't really need a declared type, its just as fast to use `Partial<Donor>` where this is required.

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+		headers: {
+			Authorization: 'Bearer ' + token,
+		},
+	};
+
+	let response;
+	try {
+		const url =
+			config.getConfig().idServiceUrl() +
+			`${req.programId}/${req.submitterDonorId}/${req.submitterSpecimenId}/${req.submitterSampleId}/${req.submitterPrimaryDiagnosisId}/${req.submitterFollowUpId}/${req.submitterTreatmentId}/${req.testInterval}/${req.family_relative_id}/${req.comorbidityTypeCode}/${req.entityType}`;
```
joneubank, line 71: Strongly encourage the use of URL building libraries for this specific purpose: https://www.npmjs.com/package/url-join

See also the JS URL constructor: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

joneubank, line 71: I don't understand why we need absolutely all of these submitter IDs to get an ID... surely there are different entities that require different submitter properties to fetch their IDs, not all part of the same query string?

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+	submitterSampleId: '-',
+	submitterPrimaryDiagnosisId: '-',
+	submitterFollowUpId: '-',
+	submitterTreatmentId: '-',
+	submitterDonorId: '-',
+	testInterval: '-',
+	family_relative_id: '-',
+	comorbidityTypeCode: '-',
+	entityType: '-',
+};
```
joneubank, line 54: Probably just want to use a Partial version of this object and then let the consuming function convert the `undefined` values to whatever palceholder string we want. This is simpler than requiring the user submitting the request to understand that `'-'` is a placeholder value.

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+
+export async function setEntityIdsForDonors(donors: Donor[]) {
+	const donorsWithIds = donors.map(async (donor) => {
+		return await setEntityIds(donor);
+	});
+	return await Promise.all(donorsWithIds);
+}
+
+export async function setEntityIds(donor: PartialDonor) {
+	const submitterDonorId = donor.submitterId as string;
```
joneubank, line 91: We don't know that `donor.submitterId` is a string. Why are we asserting this? Should we not be treating it based on the input parameter that you defined as `Partial<Donor>`?

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+	// -- DONOR --
+	const donorId = await getId({
+		...request,
```
joneubank, line 94: This relates to my previous comment about using a Partial type for input. You can do this `...request` filling inside of the `getId` function so that we don't need to remember to do this when calling the function.

src/clinical/id-generator.ts
```diff
@@ -0,0 +1,267 @@
...
+		programId: donor.programId as string,
+		submitterDonorId: donor.submitterId as string,
```
joneubank, line 96: We don't know that these are strings.

src/utils.ts
```diff
@@ -125,6 +125,10 @@ export namespace ControllerUtils {
 		return res.status(422).send({ message });
 	};
 
+	export const unableToGenerateId = (res: Response, message: string): any => {
```
joneubank, line 128: don't declare the return type as `any`, if you are unsure what return type you want then let the function infer it from your return statement.

src/clinical/donor-repo.ts
```diff
@@ -109,20 +110,22 @@ export interface DonorRepository {
...
 
 // Mongoose implementation of the DonorRepository
 export const donorDao: DonorRepository = {
 	async insertDonors(donors: Donor[]) {
-		await mongoose.connection.db.collection('donors').insertMany(donors);
+		const donorsWithIds = await setEntityIdsForDonors(donors);
+		await mongoose.connection.db.collection('donors').insertMany(donorsWithIds);
 	},
 	async updateDonor(donor: Donor) {
+		const donorsWithIds = await setEntityIds(donor);
```
joneubank, line 125: Both `setEntityIds` functions throw errors when network errors occur, should there be some sort of error handling here to handle failures or is this now expected to be handled by the caller of `updateDonor` and `insertDonors`?

joneubank (commented):
Generally workable solution. The ID path is shockingly long and I am wondering if this is a configuration/design concern with the ID service instead of a problem with the implementation here.

My only feedback for changes is to listen to the TS errors and perform required type checks instead of just asserting new types. TS can't provide us any meaningful protection against invalid types if we keep telling it that it is wrong.

src/clinical/id-generator.ts
joneubank: Discussed this offline. This aligns with the current ID Service implementation and is correct at the current time.

src/clinical/id-generator.ts
joneubank: Not repeating this comment, but this same pattern continues where the values may be undefined but are not being checked. This could result in requests to the ID service with `undefined` values in the path params.

src/clinical/id-generator.ts
justincorrigible: Dan's suggestion, for clarity
```suggestion
		console.log(`getId response: ${response.data.entityId} - ${response.data.entityType}`);
```

PR #1178, opened 2024-04-16
🏷️ 1141 Add Exception Manifest to Donor tsv

demariadaniel:
## Link to Issue

[#1141](https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/1141)

## Description

- Adds Exception Manifest to `getDonorDataByIdAsTsvsInZip` at `/donors/tsv`

- Allows downloading exceptions from multiple programs if files from multiple programs are selected (which is possible, tested in QA)
Example output in ticket

### Type of Change

- [ ] Bug
- [ ] Refactor
- [x] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/clinical/api/clinical-api.ts
+++ b/src/clinical/api/clinical-api.ts
@@ -59,22 +59,19 @@ export const parseDonorIdList = (donorIds: string) =>
 
 export const createClinicalZipFile = (
 	data: ClinicalEntityData[],
-	exceptionManifest?: {
-		programShortName: string;
-		exceptions: ExceptionManifestRecord[];
-	},
+	exceptions?: ExceptionManifestRecord[],
 ) => {
 	const zip = new AdmZip();
 	data.forEach((entityData) => {
 		const tsvData = TsvUtils.convertJsonRecordsToTsv(entityData.records, entityData.entityFields);
 		zip.addFile(`${entityData.entityName}.tsv`, Buffer.alloc(tsvData.length, tsvData));
 	});
-	if (exceptionManifest && exceptionManifest.exceptions.length) {
-		const headers = exceptionManifest.exceptions
+	if (exceptions?.length) {
+		const headers = exceptions
 			.map((exception) => Object.keys(exception))
 			.flat()
 			.filter((key, index, keyArray) => keyArray.indexOf(key) === index);
-		const tsvData = TsvUtils.convertJsonRecordsToTsv(exceptionManifest.exceptions, headers);
+		const tsvData = TsvUtils.convertJsonRecordsToTsv(exceptions, headers);
 		zip.addFile(`exceptions_manifest.tsv`, Buffer.alloc(tsvData.length, tsvData));
 	}
 	return zip;
@@ -160,7 +157,7 @@ class ClinicalController {
 			.attachment(fileName)
 			.setHeader('content-disposition', fileName);
 
-		const zip = createClinicalZipFile(entityData, { programShortName, exceptions });
+		const zip = createClinicalZipFile(entityData, exceptions);
 
 		res.send(zip.toBuffer());
 	}
@@ -188,6 +185,7 @@ class ClinicalController {
 
 		const date = currentDateFormatted();
 		const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
+
 		res
 			.status(200)
 			.contentType('application/zip')
--- a/src/clinical/api/clinical-api.ts
+++ b/src/clinical/api/clinical-api.ts
@@ -183,14 +183,30 @@ class ClinicalController {
 
 		const donorEntityData = await service.getDonorEntityData(donorIds);
 
+		const donors = await service.getDonorsByIds(donorIds);
+
+		// File Table can request multiple programs
+		const donorPrograms = donors
+			.map((donor) => donor.programId)
+			.filter((programId, index, array) => array.indexOf(programId) === index);
+
 		const date = currentDateFormatted();
 		const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
 
-		// const exceptions =
-		// 	(await getExceptionManifestRecords(programShortName, {
-		// 		donorIds,
-		// 		submitterDonorIds,
-		// 	})) || [];
+		const exceptions: ExceptionManifestRecord[][] = [];
+		for (const programShortName of donorPrograms) {
+			const programDonors = donors
+				.filter((donor) => donor.programId === programShortName)
+				.map((donor) => donor.donorId);
+
+			const programExceptions =
+				(await getExceptionManifestRecords(programShortName, {
+					donorIds: programDonors,
+					submitterDonorIds: [],
+				})) || [];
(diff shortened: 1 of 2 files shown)


src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,39 @@ class ClinicalController {
...
+		).then((exceptionsArray) => exceptionsArray.flat());
+
 		res
 			.status(200)
 			.contentType('application/zip')
 			.attachment(fileName)
 			.setHeader('content-disposition', fileName);
 
-		const zip = createClinicalZipFile(donorEntityData);
+		const zip = createClinicalZipFile(donorEntityData, exceptions);
```
joneubank, line 218: This reads very nicely now. Zip file will contain donor entity data and exceptions. Looks great.

src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,39 @@ class ClinicalController {
 
 		const donorEntityData = await service.getDonorEntityData(donorIds);
 
+		const entityRecords = donorEntityData[0].records;
+
+		// File Table can request multiple programs
+		const donorPrograms = entityRecords
+			.map((record) => record.program_id as string)
+			.filter((programId, index, array) => programId && array.indexOf(programId) === index);
```
ciaranschutte, line 191: This filter isn't doing anything.
It's always going to be `true` because you're filtering on the array already and `index` is the current iteration index.
So `arrayIndexOf(programId)` and `index` are the same value.

src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,39 @@ class ClinicalController {
...
+
 		const date = currentDateFormatted();
 		const fileName = `filename=Donor_Clinical_Data_${date}.zip`;
+
+		const exceptions = await Promise.all(
+			donorPrograms.map(async (programShortName) => {
+				const programDonors = entityRecords
+					.filter((record) => record.donor_id && record.program_id === programShortName)
+					.map((record) => record.donor_Id as number);
```
joneubank, line 200: Can you do `.map((record) => Number(record.donor_Id))` instead of the type assertion? I'm not sure what type these are coming back as and this will ensure the type is converted correctly. It will also take weird strings and make them into `NaN` so you may want a filter step after this conversion. 

Also, should I be concerned about the mixed case of `donor_id` and `donor_Id` on neighbouring lines?

src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,39 @@ class ClinicalController {
 
 		const donorEntityData = await service.getDonorEntityData(donorIds);
 
+		const entityRecords = donorEntityData[0].records;
```
joneubank, line 186: We need to check that `donorEntityData` has at least one item. `donorEntityData[0]` could be undefined.

src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,39 @@ class ClinicalController {
...
+
+				const programExceptions =
+					(await getExceptionManifestRecords(programShortName, {
+						donorIds: programDonors,
+						submitterDonorIds: [],
+					})) || [];
+
+				return programExceptions;
+			}),
+		).then((exceptionsArray) => exceptionsArray.flat());
```
joneubank, line 210: I'm going to accept this PR without these changes, but this is some pretty heavily nested code thats hard for me to parse. Imagine it reorganized like:

```ts
const getEntityExceptionsByProgramName = async (programShortName: string) => {
	const programDonors = entityRecords
		.filter((record) => record.donor_id && record.program_id === programShortName)
		.map((record) => record.donor_Id as number);
	const programExceptions =
		(await getExceptionManifestRecords(programShortName, {
			donorIds: programDonors,
			submitterDonorIds: [],
		})) || [];

	return programExceptions;
}

const programExceptionPromises = donorPrograms.map(getProgramExceptionsForDonors);

const exceptionsArray = await Promise.all(programExceptionPromises);
const exceptions = exceptionsArray.flat();

```

joneubank (changes requested):
Good improvements. I think there is one type check missing when accessing an array by index that could cause an issue so I'd like that addressed before approving.

The comments about parsing Number vs assertion, and refactoring the nested promise map are suggestions if you have time.

src/clinical/api/clinical-api.ts
joneubank: This one needs to be checked before we merge. 🙏

ciaranschutte:
I don't think any of this `as` TS usage is correct.
I understand you want to narrow the type but you're not actually checking to make sure the value is of that type.

We are telling TS already that these things are of a certain type eg. https://github.com/icgc-argo/argo-clinical/blob/develop/src/clinical/clinical-entities.ts#L115 and this usage negates that.

I believe the fix is to add explicit check if TS can't infer.
eg. if you're saying `as string` make sure `typeof foo === 'string'`

calling TS gods for 2nd opinion @joneubank

src/clinical/api/clinical-api.ts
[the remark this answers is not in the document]
demariadaniel: Oops yes you guys are right, rushed ahead without solving this properly. Let me revise.

demariadaniel: Yes sorry, updating

src/clinical/api/clinical-api.ts
ciaranschutte: great! If you fancy adding a comment I'd appreciate it, as I didn't grok it

src/clinical/api/clinical-api.ts
demariadaniel: Added numeric conversion, fixed property name

src/clinical/api/clinical-api.ts
demariadaniel: Yes async maps get very ugly ... I improved readability but used a different approach w/ a `for ... of` loop

src/clinical/api/clinical-api.ts
```diff
@@ -186,15 +183,56 @@ class ClinicalController {
...
+				);
+
+			for (const programShortName of donorPrograms) {
+				// Get Donor Ids + Exceptions for each program
+				const programDonors = entityRecords
+					.filter(
+						(record) =>
+							typeof record.donor_id === 'number' && record.program_id === programShortName,
+					)
+					.map((record) => Number(record.donor_id))
```
joneubank, line 210: This `Number()` conversion is no longer necessary.
```suggestion
					.filter(
						(record): record is (ClinicalInfo & {donor_id: number, program_id: string}) =>
							typeof record.donor_id === 'number' && record.program_id === programShortName,
					)
					.map((record) => record.donor_id)
```

joneubank (approved):
good stuff, thanks!

joneubank (approved):

PR #1184, opened 2024-04-29
❌ Sort Invalid Clinical Records first

demariadaniel:
## Link to Issue

[#2624](https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/platform-ui/2624)

## Description
Sorts Invalid Clinical Records to the top of the list

Donor records already include sorting rules for `isValid`; this change impacts Clinical Entity records

Relates to UI changes in: https://github.com/icgc-argo/platform-ui/pull/2763

Testing examples are in ticket comments

Updates Lectern Client to 1.5 so that all errors have populated values; allows accurate sorting into invalid/valid entity record groups

Screenshot 2024-05-16 at 3 08 02 PM

### Type of Change

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [ ] Manual testing

### DIFF
--- a/src/clinical/clinical-service.ts
+++ b/src/clinical/clinical-service.ts
@@ -24,6 +24,7 @@ import {
 import { DeepReadonly } from 'deep-freeze';
 import _ from 'lodash';
 import {
+	ClinicalDataSortTypes,
 	ClinicalEntityErrorRecord,
 	ClinicalEntitySchemaNames,
 	ClinicalErrorsResponseRecord,
@@ -69,7 +70,7 @@ export type PaginationQuery = {
 	sort: string;
 };
 
-type ClinicalDataPaginatedQuery = ClinicalDonorEntityQuery & PaginationQuery;
+export type ClinicalDataPaginatedQuery = ClinicalDonorEntityQuery & PaginationQuery;
 
 export type ClinicalDataQuery = ClinicalDataPaginatedQuery & {
 	completionState?: {};
@@ -231,8 +232,31 @@ export const getPaginatedClinicalData = async (programId: string, query: Clinica
 	// Get all donors + records for given entity
 	const { donors, totalDonors } = await donorDao.findByPaginatedProgramId(programId, query);
 
+	const donorIds = donors.map((donor) => donor.donorId);
+
+	const isDefaultDonorSort = query.sort.includes('completionStats.coreCompletionPercentage');
+	const isInvalidSort = query.sort.includes('schemaMetadata.isValid');
+
+	const clinicalErrors = isInvalidSort
+		? (await getClinicalErrors(programId, donorIds)).clinicalErrors
+		: [];
+
+	const sortType = isDefaultDonorSort
+		? ClinicalDataSortTypes.defaultDonor
+		: isInvalidSort
+		? ClinicalDataSortTypes.invalidEntity
+		: ClinicalDataSortTypes.columnSort;
+
 	const taskToRun = WorkerTasks.ExtractEntityDataFromDonors;
-	const taskArgs = [donors as Donor[], totalDonors, allSchemasWithFields, query.entityTypes, query];
+	const taskArgs = [
+		donors as Donor[],
+		totalDonors,
+		allSchemasWithFields,
+		query.entityTypes,
+		query,
+		sortType,
+		clinicalErrors,
+	];
 
 	// Return paginated data
 	const data = await runTaskInWorkerThread<{ clinicalEntities: ClinicalEntityData[] }>(
--- a/src/clinical/clinical-service.ts
+++ b/src/clinical/clinical-service.ts
@@ -24,6 +24,7 @@ import {
 import { DeepReadonly } from 'deep-freeze';
 import _ from 'lodash';
 import {
+	ClinicalDataSortType,
 	ClinicalDataSortTypes,
 	ClinicalEntityErrorRecord,
 	ClinicalEntitySchemaNames,
@@ -241,7 +242,7 @@ export const getPaginatedClinicalData = async (programId: string, query: Clinica
 		? (await getClinicalErrors(programId, donorIds)).clinicalErrors
 		: [];
 
-	const sortType = isDefaultDonorSort
+	const sortType: ClinicalDataSortType = isDefaultDonorSort
 		? ClinicalDataSortTypes.defaultDonor
 		: isInvalidSort
 		? ClinicalDataSortTypes.invalidEntity
--- a/src/clinical/service-worker-thread/tasks.ts
+++ b/src/clinical/service-worker-thread/tasks.ts
@@ -180,8 +180,10 @@ const mapEntityDocuments = (
 	switch (sortType) {
 		case ClinicalDataSortTypes.defaultDonor:
 			records = results.sort(sortDocs(sort, completionStats, sortDonorRecordsByCompletion));
+			break;
 		case ClinicalDataSortTypes.invalidEntity:
 			records = sortInvalidRecords(errors, results, entityName);
+			break;
 		case ClinicalDataSortTypes.columnSort:
 		default:
 			const sortKey = sort[0] === '-' ? sort.split('-')[1] : sort;
--- a/src/clinical/service-worker-thread/tasks.ts
+++ b/src/clinical/service-worker-thread/tasks.ts
@@ -113,16 +113,42 @@ const sortRecordsByColumn = (
 
 // Sort Invalid Records to Top
 const sortInvalidRecords = (
-	currentRecord: ClinicalInfo,
-	nextRecord: ClinicalInfo,
-	errorIds: Set<number>,
+	errors: ClinicalErrorsResponseRecord[],
+	records: ClinicalInfo[],
+	entityName: ClinicalEntitySchemaNames,
 ) => {
-	const currentRecordIsInvalid = errorIds.has(currentRecord.donor_id as number);
-	const nextRecordIsInvalid = errorIds.has(nextRecord.donor_id as number);
-	const invalidSort =
-		currentRecordIsInvalid === nextRecordIsInvalid ? 0 : currentRecordIsInvalid ? -1 : 1;
+	const entityErrors = errors.filter((errorRecord) => errorRecord.entityName === entityName);
+	const errorIds = new Set(entityErrors.map((error) => error.donorId));
 
-	return invalidSort;
+	const validRecords: ClinicalInfo[] = [];
+	const invalidRecords: ClinicalInfo[] = [];
+
+	records.forEach((record) => {
+		if (typeof record.donor_id === 'number') {
+			if (!errorIds.has(record.donor_id)) {
+				validRecords.push(record);
+			} else {
+				const currentRecordIsInvalid = entityErrors.find((errorRecord) => {
+					const idValid = errorRecord.donorId === record.donor_id;
+					const recordValid = errorRecord.errors.some((error) => {
+						const recordValue = record[error.fieldName];
+						const errorValue = Array.isArray(error.info.value)
+							? error.info.value[0]
+							: error.info.value;
+						return recordValue === errorValue;
+					});
+					return idValid && recordValid;
+				});
+				if (currentRecordIsInvalid) {
+					invalidRecords.push(record);
+				} else {
+					validRecords.push(record);
+				}
+			}
+		}
+	});
+
+	return [...invalidRecords, ...validRecords];
 };
 
 // Formats + Organizes Clinical Data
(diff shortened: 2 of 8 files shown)


src/clinical/clinical-service.ts
```diff
@@ -24,6 +24,7 @@ import {
 import { DeepReadonly } from 'deep-freeze';
 import _ from 'lodash';
 import {
+	ClinicalDataSortTypes,
```
demariadaniel, line 27: Due to circular imports I encountered while making this branch, I've created a second PR to move several types into their own file
It's based off of this branch, so should be reviewed after this is merged
https://github.com/icgc-argo/argo-clinical/pull/1185

src/clinical/clinical-service.ts
```diff
@@ -231,8 +232,31 @@ export const getPaginatedClinicalData = async (programId: string, query: Clinica
...
 	const { donors, totalDonors } = await donorDao.findByPaginatedProgramId(programId, query);
 
+	const donorIds = donors.map((donor) => donor.donorId);
+
+	const isDefaultDonorSort = query.sort.includes('completionStats.coreCompletionPercentage');
+	const isInvalidSort = query.sort.includes('schemaMetadata.isValid');
+
+	const clinicalErrors = isInvalidSort
+		? (await getClinicalErrors(programId, donorIds)).clinicalErrors
+		: [];
```
demariadaniel, line 242: Because Clinical Entity Data is stored on Donors, and Error records are stored on Migrations and require Dictionary/Exception validation, we have to compare Entity Records against Error Records to insure proper sorting

This duplicates the work of the Clinical Errors resolver when querying for both Clinical Data and Clinical Errors
This makes the whole operation fairly slow for large programs

I'm not sure how to optimize this without a substantial refactor. 
Open to suggestions, but improvements should be made a separate ticket.

joneubank: It requires refactoring how the data is stored. Thanks for finding a way to work with it as is and highlighting the issue.

joneubank: Additional thought: this will be very quick for programs that don't have migration errors, so as long as we can keep migration errors in check its a non-issue.

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -116,6 +118,20 @@ const sortRecordsByColumn = (
...
+	nextRecord: ClinicalInfo,
+	errors: ClinicalErrorsResponseRecord[],
+) => {
+	const currentRecordIsInvalid = errors.some((record) => record.donorId === currentRecord.donor_id);
+	const nextRecordIsInvalid = errors.some((record) => record.donorId === nextRecord.donor_id);
+	const invalidSort =
+		currentRecordIsInvalid === nextRecordIsInvalid ? 0 : currentRecordIsInvalid ? -1 : 1;
+
+	return invalidSort;
+};
```
joneubank, line 133: This looks logically correct, but I can see how this is very slow. We need to iterate the errors list twice for every element in our array. If before iterating over your array you prepare a map of donorId to validity state and pass that as an argument to the sort function then you can just look up the validity of each donorId instead of checking the errors array.

```ts
const sortByValidity = (errors: ClinicalErrorsResponseRecord[]) => {
	const donorIdsWithErrors = errors.reduce<Set<any>>((acc, error) => acc.add(error.donorId), new Set());
	return (currentRecord: ClinicalInfo, nextRecord: ClinicalInfo): number => {
		const currentRecordIsInvalid = donorIdsWithErrors.has(currentRecord.donor_id);
		const nextRecordIsInvalid = donorIdsWithErrors.has(nextRecord.donor_id);
		return currentRecordIsInvalid === nextRecordIsInvalid ? 0 : currentRecordIsInvalid ? -1 : 1;
	};
};
```

Another way to improve this is to build 2 arrays, one of the valid donors and one of the invalid donors, sort each of those arrays, and then combine them. This way you don't need to use a compareFunction to sort valid vs invalid.

joneubank (commented):
Can you take one more pass over this code to see if theres a better way to structure the different query requests? Your general point about how we need to do awkward looping over the errors to map them to our donors is legitimate, that is an unfortunate consequence of the data model. The part that seems like it needs some revision is the function signature that requires `errors` to be provided to `extractEntityDataFromDonors` and then `mapEntityDocuments` even if the `sortType` is not set to `invalidDonors`.

I also provided an example function that reduces the operations for sorting an invalid donor from O(N^2) to O(N) - it only needs to loop over the error array once, it doesn't need to perform a find operation on the error array for every record in the donor array. This could help with your performance concerns

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -58,21 +60,21 @@ const isEntityInQuery = (entityName: ClinicalEntitySchemaNames, entityTypes: str
 // Main Sort Function
 const sortDocs = (
 	sortQuery: string,
-	entityName: string,
 	completionStats: CompletionDisplayRecord[],
+	sortType: keyof typeof ClinicalDataSortTypes,
```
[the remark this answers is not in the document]
demariadaniel: So this second point is worth discussing in depth. I don't think it's well defined what behaviour we want.

I approached this as, invalid records should be prioritized first by default until another sorting option is selected. You're seeing it as, invalid records should be prioritized first in every scenario.

My thinking is, if invalid records are always prioritized, and a User selects a specific Column to sort by, they aren't necessarily seeing the 'last' or 'highest' value at the top of that column. It disrupts the logical 'A-Z, 1-10' you'd expect to some degree. Maybe that's fine though, because error records will be sorted.


I solved this in such a way that invalid records are prioritized on first page load. They are also sorted by DonorId, so there is a combination.
But if a User chooses a specific sorting column, there's no way to go back to 'invalid record sorting' other than a page refresh.

Either way I don't think it's defined which behaviour is expected. Both have tradeoffs.

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -124,20 +140,23 @@ const mapEntityDocuments = (
 	entityTypes: EntityAlias[],
 	paginationQuery: PaginationQuery,
 	completionStats: CompletionDisplayRecord[],
+	sortType: keyof typeof ClinicalDataSortTypes,
+	errors: ClinicalErrorsResponseRecord[],
```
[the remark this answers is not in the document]
demariadaniel: I can stream line this, but I did make sure that errors are only calculated when needed: https://github.com/icgc-argo/argo-clinical/blob/bug/2624-sort-invalid-clinical-records/src/clinical/clinical-service.ts#L240

src/clinical/service-worker-thread/tasks.ts
demariadaniel: OK I've updated this to use Sets and avoid multiple iterations, it's definitely a little faster now too
https://github.com/icgc-argo/argo-clinical/pull/1184/commits/4281c6a3d7b2f086d68369f3920e43badc8bbf53

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -116,6 +108,20 @@ const sortRecordsByColumn = (
...
 };
 
+// Sort Invalid Records to Top
+const sortInvalidRecords = (
+	currentRecord: ClinicalInfo,
+	nextRecord: ClinicalInfo,
+	errorIds: Set<number>,
+) => {
+	const currentRecordIsInvalid = errorIds.has(currentRecord.donor_id as number);
+	const nextRecordIsInvalid = errorIds.has(nextRecord.donor_id as number);
```
joneubank, line 118: Leaving comment to state I reviewed this type assertion and think its ok in this case. If the donor_id is not a number then the `has` function will fail to match it pretty safely.

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -57,27 +57,17 @@ const DONOR_ID_FIELD = 'donor_id';
...
 	entityTypes.includes(aliasEntityNames[entityName]);
 
-// Main Sort Function
+// Base Sort Function Wrapper
 const sortDocs = (
 	sortQuery: string,
-	completionStats: CompletionDisplayRecord[],
-	sortType: keyof typeof ClinicalDataSortTypes,
-	errors: ClinicalErrorsResponseRecord[],
+	sortFunction: (currentRecord: ClinicalInfo, nextRecord: ClinicalInfo, rest: any) => number,
```
joneubank, line 63: I think the type on `rest` could be more specific.

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -149,14 +139,33 @@ const mapEntityDocuments = (
...
+		sortType === ClinicalDataSortTypes.invalidEntity
+			? entityErrors.reduce<Set<number>>((acc, error) => acc.add(error.donorId), new Set())
+			: new Set([]);
+
+	const sortRecords =
+		sortType === ClinicalDataSortTypes.defaultDonor
+			? sortDocs(sort, sortDonorRecordsByCompletion, completionStats)
+			: sortType === ClinicalDataSortTypes.invalidEntity
+			? sortDocs(sort, sortInvalidRecords, donorIdsWithErrors)
+			: sortDocs(sort, sortRecordsByColumn, key);
```
joneubank, line 166: This is so hard to follow with all the ternaries... could we instead just do a switch on the different sort types and have the logic for each contained in its own block?

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -57,27 +57,17 @@ const DONOR_ID_FIELD = 'donor_id';
...
 
-// Main Sort Function
+// Base Sort Function Wrapper
 const sortDocs = (
 	sortQuery: string,
-	completionStats: CompletionDisplayRecord[],
-	sortType: keyof typeof ClinicalDataSortTypes,
-	errors: ClinicalErrorsResponseRecord[],
+	sortFunction: (currentRecord: ClinicalInfo, nextRecord: ClinicalInfo, rest: any) => number,
+	sortArgs: any,
```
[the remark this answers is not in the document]
demariadaniel: Sorry that was rushed, this is now updated

src/clinical/service-worker-thread/tasks.ts
demariadaniel: Switch/case solution w/ TS generics to solve some of the issues noted above
https://github.com/icgc-argo/argo-clinical/pull/1184/commits/532c648c059826a5e35ab25ef8df8dcf40841307

src/common-model/entities.ts
```diff
@@ -94,6 +94,12 @@ export interface ClinicalErrorsResponseRecord {
 	errors: ClinicalEntityErrorRecord[];
 }
 
+export const ClinicalDataSortTypes = {
+	defaultDonor: 'defaultDonor',
+	invalidEntity: 'invalidEntity',
+	columnSort: 'columnSort',
+};
```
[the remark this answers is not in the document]
demariadaniel: Updated

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -137,7 +148,25 @@ const mapEntityDocuments = (
...
+
+	switch (sortType) {
+		case ClinicalDataSortTypes.defaultDonor:
+			records = results.sort(sortDocs(sort, completionStats, sortDonorRecordsByCompletion));
+		case ClinicalDataSortTypes.invalidEntity:
+			const entityErrors = errors.filter((errorRecord) => errorRecord.entityName === entityName);
+			const donorIdsWithErrors = entityErrors.reduce<Set<number>>(
+				(acc, error) => acc.add(error.donorId),
+				new Set(),
+			);
```
ciaranschutte, line 162:
```js
const donorIdsWithErrors = new Set(entityErrors.map(error => error.donorId))
```
minor suggestion - use map with `Set` constructor. using `reduce` and not reducing to a value is confusing.

demariadaniel: Tested and Set prevents duplicate values using this syntax so it works just as well as using reduce/add

src/clinical/service-worker-thread/tasks.ts
```diff
@@ -180,8 +180,10 @@ const mapEntityDocuments = (
 	switch (sortType) {
 		case ClinicalDataSortTypes.defaultDonor:
 			records = results.sort(sortDocs(sort, completionStats, sortDonorRecordsByCompletion));
+			break;
 		case ClinicalDataSortTypes.invalidEntity:
 			records = sortInvalidRecords(errors, results, entityName);
+			break;
 		case ClinicalDataSortTypes.columnSort:
 		default:
 			const sortKey = sort[0] === '-' ? sort.split('-')[1] : sort;
```
joneubank, line 189: Appreciate you adding the break statements.

Other things that I would reccommend are:
- use brackets to define a scoped block for each case
- add a comment after the `columnSort` case stating that you are intentionally falling through to default

joneubank (approved):
I think we've gotten this to be good enough within the current setup, without refactoring the tasks or the stored migration error data. Thanks!

src/clinical/service-worker-thread/tasks.ts
joneubank: That's fine. These are stylistic things not functional.

PR #1187, opened 2024-05-07
fix to resolve incorrect merging of therapies into treatments

UmmulkiramR:
The code will now merge treatment records before merging therapies.

### DIFF
--- a/src/submission/submission-to-clinical/merge-submission.ts
+++ b/src/submission/submission-to-clinical/merge-submission.ts
@@ -60,7 +60,22 @@ export const mergeActiveSubmissionWithDonors = async (
 	donors: readonly DeepReadonly<Donor>[],
 ) => {
 	const updatedDonors = _.cloneDeep(donors as Donor[]);
-	for (const entityType in activeSubmission.clinicalEntities) {
+	mergeClinicalEntities(activeSubmission, updatedDonors);
+	mergeTherapiesInTreatments(activeSubmission, updatedDonors);
+	return updatedDonors;
+};
+
+const mergeClinicalEntities = (
+	activeSubmission: DeepReadonly<ActiveClinicalSubmission>,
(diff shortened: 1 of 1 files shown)


src/submission/submission-to-clinical/merge-submission.ts
```diff
@@ -126,9 +136,26 @@ export const mergeActiveSubmissionWithDonors = async (
...
+) => {
+	ClinicalTherapySchemaNames.forEach((therapyType) => {
+		const entityData = activeSubmission.clinicalEntities[therapyType];
+		entityData?.records.forEach((record) => {
+			const donorId = record[DonorFieldsEnum.submitter_donor_id];
+			const donor = _.find(updatedDonors, ['submitterId', donorId]);
+			if (!donor) {
+				throw new Errors.StateConflict(
+					`Donor ${donorId} has not been registered but is part of the activeSubmission, merge cannot be completed.`,
+				);
```
joneubank, line 154: Throwing errors from a loop. Do we want the entire process to break, or just skip this step?

I also notice the invoking method does not have a try catch. Will it stop the entire data submission process? What state will it be left in, will it be partially changed and saved?

At very least, we need to log what issue has occurred with some not about the expected consequences and what actions (if any) are needed.

UmmulkiramR: To ensure there is minimal impact on the existing functioning of the code, all this fix does is break up one big method into 2 parts ensuring the treatments are processed before the treatment types. Any error handling that was a part of the original big function has been taken exactly in the 2 new functions. The error being thrown from within the loop was how it was earlier and that wasn't changed to avoid inducing new bugs and ensure all error scenarios are handled exactly the way it was handled before.

joneubank (approved):

PR #1214, opened 2024-11-06
Feat/#1186 disable cross field validation for survival time

ciaranschutte:
## Link to Issue

https://github.com/icgc-argo/roadmap/issues/1186

## Description

disables time conflict validation on follow_up, specimen and treatment if exception exists

### Type of Change

- [x] Bug
- [ ] Refactor
- [ ] New Feature
- [ ] Release Candidate

- [x] Manual testing

### DIFF
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -345,18 +345,20 @@ export async function getExceptionManifestRecords(
  * @param schema
  * @returns boolean indicating if an exception exists
  */
-export const checkForExceptions = async (
-	record: DeepReadonly<SubmittedClinicalRecord>,
-	field: string,
-	schema: string,
-): Promise<boolean> => {
+export const checkForExceptions = async ({
+	record,
+	field,
+	schema,
+}: {
+	record: DeepReadonly<SubmittedClinicalRecord>;
+	field: string;
+	schema: string;
+}): Promise<boolean> => {
 	const programId = record['program_id'] as string;
 
 	if (!programId) return false;
 
-	const programAdditionalSearchParams = {
-		exceptions: { requested_core_field: field },
-	};
+	const programAdditionalSearchParams = { requested_core_field: field };
 
 	const entityAdditionalSearchParams = {
 		[`${schema}.requested_core_field`]: field,
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -335,3 +335,23 @@ export async function getExceptionManifestRecords(
 
 	return exceptionManifest;
 }
+
+/**
+ * Boolean check for existence of exceptions for a specific field
+ *
+ * @param record
+ * @param field
+ * @returns boolean indicating if an exception exists
+ */
+export const checkForExceptions = async (record: any, field: any): Promise<boolean> => {
+	const programId = record['program_id'];
+	const additionalSearchParams = {
+		exceptions: { requested_core_field: field },
+	};
+
+	// retrieve submitted exceptions for program id (both program level and entity level)
+	const programException = await programExceptionRepository.find(programId, additionalSearchParams);
+	//const entityException = await entityExceptionRepository.find(programId);
+
+	return notEmpty(programException); // || notEmpty(entityException);
+};
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -44,6 +44,7 @@ import {
 	findDonorsBySubmitterIds,
 } from '../../clinical/clinical-service';
 import { notEmpty } from '../../utils';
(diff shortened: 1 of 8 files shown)


src/submission/validation-clinical/followUp.ts
```diff
@@ -24,13 +24,82 @@ import {
 	SubmissionValidationOutput,
 	DonorVitalStatusValues,
 } from '../submission-entities';
-import { ClinicalEntitySchemaNames, FollowupFieldsEnum } from '../../common-model/entities';
+import {
```
ciaranschutte, line 27: changes here mirror changes to other validation files.
big diff hash because I moved an inline function out of a function body

src/submission/validation-clinical/followUp.ts
```diff
@@ -24,13 +24,82 @@ import {
...
+	treatment: DeepReadonly<Treatment>,
+	errors: SubmissionValidationError[],
+) {
+	// A follow up may or may not be associated with treatment
+	if (treatment == undefined) return;
+
+	if (
+		followUpRecord.interval_of_followup &&
+		treatment.clinicalInfo &&
+		treatment.clinicalInfo.treatment_start_interval &&
```
demariadaniel, line 72: Potential dedupe with `treatment.clinicalInfo?.treatment_start_interval`

src/submission/validation-clinical/followUp.ts
```diff
@@ -24,13 +24,82 @@ import {
...
+function getExistingFollowUp(
+	existingDonor: DeepReadonly<Donor>,
+	record: DeepReadonly<SubmittedClinicalRecord>,
+) {
+	if (existingDonor.followUps) {
+		return getClinicalEntitiesFromDonorBySchemaName(
+			existingDonor,
+			ClinicalEntitySchemaNames.FOLLOW_UP,
+		).find(
+			(ci) =>
```
demariadaniel, line 95: Legacy function but opportunity to improve variable name `ci` here

src/submission/exceptions/exceptions.ts
```diff
@@ -335,3 +336,39 @@ export async function getExceptionManifestRecords(
...
+ * @param field
+ * @param schema
+ * @returns boolean indicating if an exception exists
+ */
+export const checkForExceptions = async (
+	record: DeepReadonly<SubmittedClinicalRecord>,
+	field: string,
+	schema: string,
+): Promise<boolean> => {
+	const programId = record['program_id'] as string;
```
demariadaniel, line 353: Error handling if `program_id` is `undefined`? 
`SubmittedClinicalRecord` type doesn't enforce that key so I'm wondering if there's potential for error here

ciaranschutte:
CI failing due to tests. https://github.com/icgc-argo/argo-clinical/pull/1216 should fix it.

src/submission/validation-clinical/followUp.ts
ciaranschutte: Will leave as is for now, lots of cases to cleanup with many conditions relying on this check (not just line 72)

src/submission/exceptions/exceptions.ts
ciaranschutte: added handling

src/submission/validation-clinical/followUp.ts
ciaranschutte: Good call. Will leave as is for now because this is just a bug fix PR rather than a chore or feature in our terminology

joneubank (commented):
Here are my comments about the types used in the code provided, this is mostly about having your functions be more strict/opinionated about their interface/parameters.

joneubank:
I am proposing a change to the expected behaviour detailed in the ticket. Instead of having our cross file validations check for the `survival_time` exception, we should just check if there is a `survival_time` value. If there is no value, then we don't apply the checks (they would always fail). If there is no value and also no exception, then the validation would fail in the donor entity validation, so we don't need to worry about that in the time interval cross validations.

I'll add a comment to the issue to the same effect and create a new PR into this branch with the proposed changeds.

ciaranschutte:
tested locally.

src/submission/validation-clinical/utils.ts
```diff
@@ -516,44 +516,31 @@ export function getRelatedEntityByFK(
...
-			missingDonorFields.push(DonorFieldsEnum.survival_time);
+			return undefined;
 		}
 	}
 
-	if (missingDonorFields.length > 0) {
-		errors.push(
-			buildSubmissionError(
-				submittedRecord,
-				DataValidationErrors.NOT_ENOUGH_INFO_TO_VALIDATE,
```
demariadaniel, line 548: Source of testing errors -- this is removed and not replaced elsewhere. The tests are looking for the `NOT_ENOUGH_INFO_TO_VALIDATE` error and it's not found.

joneubank (approved):

PR #1226, opened 2025-05-21
QA feedback - single specimen exception

ciaranschutte:
QA feedback - single specimen exception
Parse result object correctly

### DIFF
--- a/src/clinical/clinical-service.ts
+++ b/src/clinical/clinical-service.ts
@@ -200,6 +200,15 @@ export const getClinicalData = async (programId: string) => {
 	return data;
 };
 
+const getSingleSpecimenExceptions = async () => {
+	const result = await getExceptions();
+	if (result.success) {
+		return result.data;
+	} else {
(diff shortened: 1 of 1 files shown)


src/clinical/clinical-service.ts
```diff
@@ -200,6 +200,15 @@ export const getClinicalData = async (programId: string) => {
 	return data;
 };
 
+const getSingleSpecimenExceptions = async () => {
+	const result = await getExceptions();
+	if (result.success) {
+		return result.data;
+	} else {
+		throw Error('Database failure retrieving single specimen exceptions.');
+	}
```
joneubank, line 209: Do you know where this thrown exception gets handled?

joneubank (approved):

src/clinical/clinical-service.ts
ciaranschutte: global top level, similar to the other api calls

PR #1237, opened 2025-10-17
Clinical Upload single rxnorm db query optimise

ciaranschutte:
- rxnorm db queries were not being batched which causes very long processing times due to db connection limits (eg. 6000 records waiting on a pool of 10 connections)
- makes 1 DB query for all rxcui lookup values at the start of validation if applicable
- remove old service db query, it was only in use in this instance of clinical upload
- uses cache in the same pattern as exceptions by passing it as an arg
- kept some logs around but using application debugger, they will only show under "debug" log level

### DIFF
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -47,11 +47,11 @@ import { notEmpty } from '../../utils';
 import { SubmittedClinicalRecord } from '../submission-entities';
 
 /**
- * query db for program or entity exceptions
+ * query db for program and entity exceptions
  * @param programId
  * @returns program and donor level exceptions for this programId
  */
-const queryForExceptions = async (programId: string) => {
+export const queryForExceptions = async (programId: string) => {
 	const programException = await programExceptionRepository.find(programId);
 	const entityException = await entityExceptionRepository.find(programId);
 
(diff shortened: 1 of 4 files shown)


src/submission/submission-service.ts
```diff
@@ -872,6 +872,27 @@ export namespace operations {
 		let errorsAccumulator: DeepReadonly<SubmissionValidationError[]> = [];
 		const validRecordsAccumulator: any[] = [];
 
+		let rxNormCache: Map<string, RxNormConcept[]> | undefined;
+		if (isRxNormTherapy(command.clinicalType)) {
+			const rxNormIds = new Set<string>();
```
joneubank, line 881: Thank you for deduping the list with this Set

joneubank (dismissed):

src/submission/exceptions/exceptions.ts
```diff
@@ -193,28 +193,43 @@ const isValidNumericExceptionType = (
...
 	const filteredErrors: dictionaryEntities.SchemaValidationError[] = [];
 	let normalizedRecord = record;
 
-	// retrieve submitted exceptions for program id (both program level and entity level)
-	const { programException, entityException } = await queryForExceptions(programId);
+	/*
+	 * retrieve submitted exceptions for program id (both program level and entity level)
+	 *
+	 * if exceptionsCache is provided, reads from it (no write)
+	 * else, queries exception db using programId
```
joneubank, line 223: This description would be useful in the function docs. Good enough to approve this.

joneubank (approved):

PR #1238, opened 2025-10-17
Exceptions db query optimise

ciaranschutte:
! Merging into https://github.com/icgc-argo/argo-clinical/pull/1237 other db optimise query
- clinical was querying for every record to two repos (eg. 6000 records => 1200 db calls)
- adds one db call because it only relies on programId, instead of calling for each record
- added optional cache and changes the programId to optional for the `checkForProgramAndEntityExceptions` method, as it's used elsewhere in the clinical service without a cache

### DIFF
--- a/src/submission/exceptions/exceptions.ts
+++ b/src/submission/exceptions/exceptions.ts
@@ -47,7 +47,7 @@ import { notEmpty } from '../../utils';
 import { SubmittedClinicalRecord } from '../submission-entities';
 
 /**
- * query db for program or entity exceptions
+ * query db for program and entity exceptions
  * @param programId
  * @returns program and donor level exceptions for this programId
  */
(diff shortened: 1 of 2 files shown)


src/submission/exceptions/exceptions.ts
```diff
@@ -200,21 +200,30 @@ export const checkForProgramAndEntityExceptions = async ({
...
-	programId: string;
+	programId?: string;
 	record: DeepReadonly<TypedDataRecord>;
 	schemaName: ClinicalEntitySchemaNames;
 	entitySchema: dictionaryEntities.SchemaDefinition | undefined;
 	validationErrors: dictionaryEntities.SchemaValidationError[];
+	exceptionsCache?: {
+		programException: ProgramException | null;
+		entityException: EntityException | null;
+	};
```
joneubank, line 214: Is this modifying the optional cache, or just reading from it?

I think some function documentation is needed for this change.

joneubank (changes requested):

src/submission/exceptions/exceptions.ts
ciaranschutte: https://github.com/icgc-argo/argo-clinical/pull/1238/files#diff-595129065f8b9a1620d20d77ef334b846f1f4ec8eb06bcf8f3b4de5de599e9a4R219-R224

